PR #4, opened 2017-12-08
Add tokens with git-crypt

sttts:
https://github.com/AGWA/git-crypt


sttts:
@caesarxuchao updated the key. Sgty?

PR #7, opened 2017-12-09
Makefile: wait for old pod to terminate

sttts:


Makefile
```diff
@@ -35,8 +37,9 @@ update-deps:
 .PHONY: update-deps
 
 init-deploy:
-	$(KUBECTL) delete -n "$(NAMESPACE)" rc publisher || true
-	$(KUBECTL) delete -n "$(NAMESPACE)" pod publisher || true
+	$(KUBECTL) delete -n "$(NAMESPACE)" --ignore-not-found=true rc publisher
+	$(KUBECTL) delete -n "$(NAMESPACE)" --ignore-not-found=true pod publisher
+	while $(KUBECTL) get pod publisher -a &>/dev/null; do echo -n .; sleep 1; done
```
mfojtik: don't you want to also wait for the rc to be gone?

sttts: Good question. It goes away much faster. So I never had an issue.

PR #8, opened 2017-12-09
Add config for github.com/kubernetes target-org

sttts:
This is the "production" config for the bot pushing directly into github.com/kubernetes.


sttts:
@caesarxuchao do you have anything that we have to do or add before enabling this? I would do so early next week if you agree.

configs/kubernetes-nightly-configmap.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
 data:
   config: |
     target-org: kubernetes-nightly
-    github-issue: 56876
-    dry-run: false
\ No newline at end of file
+    # github-issue: 56876
```
caesarxuchao: This was to test if the issue reporter was working, right? Can we delete it now?

configs/kubernetes-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
+metadata:
+  name: publisher-config
+data:
+  config: |
+    target-org: kubernetes
+    github-issue: 56876
```
caesarxuchao: Remove it?

caesarxuchao:
> @caesarxuchao do you have anything that we have to do or add before enabling this?

No, I'm excited to see the bot get resurrected!

Though I'll be OOO until the new year, so I won't be able to help solving issues if there is any.

configs/kubernetes-nightly-configmap.yaml
sttts: Yes, the nightly config will go away.

configs/kubernetes-configmap.yaml
sttts: This is the issue to post logs to

sttts:
> Though I'll be OOO until the new year, so I won't be able to help solving issues if there is any.

I am available most of the time for urgent issues. Moreover, we have @munnerz and @marun to get involved :)

caesarxuchao:
lgtm

PR #14, opened 2018-01-05
WIP: Make publisher-bot usable for non-kube repos

mfojtik:
This pull attempts to remove all hardcoded kube variables which should allow to use the publisher bot for repositories other than kubernetes/kubernetes.
The defaulting should be preserved, so if SourceRepo is not specified it will default to "kubernetes", thus no changes in how the publisher bot is currently deployed are needed.


cmd/publishing-bot/publisher.go
```diff
@@ -50,8 +50,7 @@ type branchRule struct {
 
 // a collection of publishing rules for a single destination repo
 type repoRules struct {
-	dstRepo string
-	// publisher.go has assumption that src.repo is always kubernetes.
```
mfojtik: @sttts not sure if I understand this assumption :-)

cmd/publishing-bot/publisher.go
```diff
@@ -65,27 +64,24 @@ type PublisherMunger struct {
...
 }
 
 // New will create a new munger.
 func New(config *github.Config) (*PublisherMunger, error) {
-	// validate config
-	if config.SourceRepo != "kubernetes" {
-		// TODO: externalize the rules below to make it generic
-		return nil, fmt.Errorf("only kubernetes as source repo supported right now")
-	}
-	if config.TargetOrg == "" {
```
mfojtik: validation moved to main.go because we need to pass the issue URL to healthz...

cmd/publishing-bot/publisher.go
sttts: There was some initial work to abstract the org, but never finished.

cmd/publishing-bot/main.go
```diff
@@ -117,10 +122,10 @@ func main() {
 			if err != nil {
 				glog.Infof("Failed to run publisher: %v", err)
 				// TODO: support other orgs
```
sttts: comment can go

cmd/publishing-bot/publisher.go
```diff
@@ -65,27 +64,24 @@ type PublisherMunger struct {
...
-	}
-
 	// create munger
 	p := &PublisherMunger{}
+
+	// set the baseRepoPath
 	gopath := os.Getenv("GOPATH")
-	p.k8sIOPath = filepath.Join(gopath, "src", "k8s.io")
+	// if the SourceRepo is not kubernetes, use github.com as baseRepoPath
+	p.baseRepoPath = filepath.Join(gopath, "src", "github.com")
```
sttts: can't we use the whole base url everywhere, e.g. k8s.io or github.com/openshift ?

sttts: for the github issue we need some special logic then of course.

cmd/collapsed-kube-commit-mapper/main.go
```diff
@@ -74,6 +74,10 @@ func main() {
 	upstreamBranch := flag.String("upstream-branch", "", "the k8s.io/kubernetes branch (fully qualified e.g. refs/remotes/origin/master) used as the filter-branch basis")
 	showMessage := flag.Bool("l", false, "list the commit message after the two hashes")
 
+	// repository flags used when the repository is not k8s.io/kubernetes
+	repoName := flag.String("repo-name", "kubernetes", "the name of the source repository")
```
sttts: would make these mandatory

mfojtik: validate if not empty you mean? are you ok with defaulting? in that case we don't have to change anything

cmd/publishing-bot/publisher.go
mfojtik: just realized this must be `filepath.Join(gopath, "src", "github.com", config.TargetOrg)`

cmd/collapsed-kube-commit-mapper/main.go
sttts: cannot forsee how many changes this really needs. But we have to pass through the non-kubernetes repo anyway. So let's use the same code path(s) for the kube case. I.e. no defaulting, but an error message if empty.

cmd/collapsed-kube-commit-mapper/main.go
```diff
@@ -74,6 +74,14 @@ func main() {
 	upstreamBranch := flag.String("upstream-branch", "", "the k8s.io/kubernetes branch (fully qualified e.g. refs/remotes/origin/master) used as the filter-branch basis")
 	showMessage := flag.Bool("l", false, "list the commit message after the two hashes")
 
+	// repository flags used when the repository is not k8s.io/kubernetes
```
sttts: outdated comment

cmd/collapsed-kube-commit-mapper/main.go
```diff
@@ -117,7 +125,7 @@ func main() {
 		glog.Fatalf("Failed to get first-parent commit list for %s: %v", dstHead.Hash, err)
 	}
 
-	kubeCommitsToDstCommits, err := git.KubeCommitsToDstCommits(r, dstFirstParents, kFirstParents)
+	kubeCommitsToDstCommits, err := git.KubeCommitsToDstCommits(r, *repoOrg, *repoName, dstFirstParents, kFirstParents)
```
sttts: KubeCommitToDstCommits -> UpstreamCommitToDstCommits

cmd/publishing-bot/healthz.go
```diff
@@ -78,7 +81,7 @@ func (h *Healthz) handler(w http.ResponseWriter, r *http.Request) {
 	h.mutex.RLock()
 	resp := h.response
 	if h.Issue != 0 {
-		resp.Issue = fmt.Sprintf("https://github.com/kubernetes/kubernetes/issues/%d", h.Issue)
+		resp.Issue = fmt.Sprintf("https://github.com/%s/%s/issues/%d", h.config.TargetOrg, h.config.SourceRepo, h.Issue)
```
sttts: maybe a comment why we choose the target org, not the source org (I agree, this is correct for ppl vendoring a foreign project).

cmd/publishing-bot/publisher.go
```diff
@@ -65,27 +64,24 @@ type PublisherMunger struct {
...
 	// create munger
 	p := &PublisherMunger{}
+
+	// set the baseRepoPath
 	gopath := os.Getenv("GOPATH")
-	p.k8sIOPath = filepath.Join(gopath, "src", "k8s.io")
+	// if the SourceRepo is not kubernetes, use github.com as baseRepoPath
+	p.baseRepoPath = filepath.Join(gopath, "src", "github.com", config.TargetOrg)
+	if config.SourceRepo == "kubernetes" {
+		p.baseRepoPath = filepath.Join(gopath, "src", "k8s.io")
```
sttts: this could be generalized. But I am fine with leaving it as it is for now.

cmd/publishing-bot/publisher.go
```diff
@@ -610,14 +608,14 @@ func (p *PublisherMunger) ensureCloned(dst string, dstURL string) error {
 
 // constructs all the repos, but does not push the changes to remotes.
 func (p *PublisherMunger) construct() error {
-	kubernetesRemote := filepath.Join(p.k8sIOPath, "kubernetes", ".git")
+	baseRemote := filepath.Join(p.baseRepoPath, p.config.SourceRepo, ".git")
```
sttts: should we call this `source*` everywhere instead of `base*`, `upstream*`, `kubernetes*`? This leads to confusion.

cmd/collapsed-kube-commit-mapper/main.go
sttts: or better SourceCommitToDstCommits. Wdyt?

cmd/sync-tags/main.go
```diff
@@ -57,9 +57,17 @@ func main() {
 	publishBranch := flag.String("branch", "", "a (not qualified) branch name")
 	prefix := flag.String("prefix", "kubernetes-", "a string to put in front of upstream tags")
 	pushScriptPath := flag.String("push-script", "", "git-push command(s) are appended to this file to push the new tags to the origin remote")
+	// repository flags used when the repository is not k8s.io/kubernetes
+	repoName := flag.String("repo-name", "", "the name of the source repository")
```
sttts: `..., e.g. kubernetes`

pkg/git/kube.go
```diff
@@ -17,23 +17,22 @@ limitations under the License.
...
-const (
-	KubernetesCommitPrefix         = "Kubernetes-commit: "
-	ancientSyncCommitSubjectPrefix = "sync(k8s.io/kubernetes)"
-)
-
-// kubeHash extracts kube commit from commit message
-func KubeHash(c *object.Commit) plumbing.Hash {
+// KubeHash extracts kube commit from commit message
+// The baseRepoName default to "kubernetes".
+func KubeHash(c *object.Commit, baseRepoOrg, baseRepoName string) plumbing.Hash {
```
sttts: SourceHash

pkg/git/kube.go
```diff
@@ -17,23 +17,22 @@ limitations under the License.
...
 	lines := strings.Split(c.Message, "\n")
+	kubernetesCommitPrefix := strings.Title(baseRepoName) + "-commit: "
```
sttts: sourceCommitPrefix

cmd/collapsed-kube-commit-mapper/main.go
mfojtik: later sound fine to me

mfojtik:
@sttts comments addressed

artifacts/scripts/publish_template.sh
```diff
@@ -89,6 +89,7 @@ EXTRA_ARGS=()
 PUSH_SCRIPT=../push-tags-${REPO}-${DST_BRANCH}.sh
 echo "#!/bin/bash" > ${PUSH_SCRIPT}
 chmod +x ${PUSH_SCRIPT}
-/sync-tags --upstream-remote upstream-kube --upstream-branch "${SRC_BRANCH}" \
+/sync-tags --repo-org kubernetes --repo-name kubernetes \
```
sttts: we need the right values here, don't we?

artifacts/scripts/util.sh
```diff
@@ -479,7 +479,7 @@ sync_repo() {
     local repo=$(basename ${PWD})
     if [ -n "$(git log --oneline --first-parent --merges | head -n 1)" ]; then
         echo "Writing k8s.io/kubernetes commit lookup table to ../kube-commits-${repo}-${dst_branch}"
-        /collapsed-kube-commit-mapper --upstream-branch refs/heads/upstream-branch > ../kube-commits-${repo}-${dst_branch}
+        /collapsed-kube-commit-mapper --repo-name kubernetes --repo-org kubernetes --upstream-branch refs/heads/upstream-branch > ../kube-commits-${repo}-${dst_branch}
```
sttts: same here.

artifacts/scripts/publish_template.sh
mfojtik: fixed.

artifacts/scripts/publish_api.sh
```diff
@@ -35,9 +35,9 @@ set -o nounset
...
 
 if [ ! $# -eq 4 ]; then
-    echo "usage: $0 src_branch dst_branch dependent_k8s_repos kubernetes_remote"
+    echo "usage: $0 src_branch dst_branch dependent_k8s_repos kubernetes_remote subdirectory source_repo_org source_repo_name" 
     exit 1
 fi
 
 SCRIPT_DIR=$(dirname "${BASH_SOURCE}")
-"${SCRIPT_DIR}"/publish_template.sh "api" "${1}" "${2}" "${3}" "${4}" "true"
+"${SCRIPT_DIR}"/publish_template.sh "api" "${1}" "${2}" "${3}" "${4}" "${5}" "${6}" "${7}" "true"
```
sttts: Note to everybody else: we plan to remove the publish scripts in a follow-up and only move the smoke tests into the rules struct. This ugly argument list will go away.

artifacts/scripts/util.sh
```diff
@@ -147,12 +149,12 @@ sync_repo() {
...
-    git remote add upstream-kube "${kubernetes_remote}" >/dev/null
-    git fetch -q upstream-kube --no-tags
+    git remote rm upstream >/dev/null || true
+    git remote add upstream "${kubernetes_remote}" >/dev/null
+    git fetch -q upstream --no-tags
     git branch -D filtered-branch >/dev/null || true
-    git branch -f upstream-branch upstream-kube/"${src_branch}"
-    echo "Checked out k8s.io/kubernetes commit $(git rev-parse upstream-branch)."
+    git branch -f upstream-branch upstream/"${src_branch}"
+    echo "Checked out source commit $(git rev-parse upstream-branch)."
```
sttts: our mixture of "source" and "upstream" is confusing. It's the same thing.

mfojtik: yeah, it is hard to formulate a good message here

artifacts/scripts/publish_api.sh
mfojtik: i should actually start an issue to track follow ups

pkg/git/kube.go
```diff
@@ -17,23 +17,22 @@ limitations under the License.
...
-)
-
-// kubeHash extracts kube commit from commit message
-func KubeHash(c *object.Commit) plumbing.Hash {
+// SourceHash extracts kube commit from commit message
+// The baseRepoName default to "kubernetes".
+func SourceHash(c *object.Commit, baseRepoOrg, baseRepoName string) plumbing.Hash {
 	lines := strings.Split(c.Message, "\n")
+	sourceCommitPrefix := strings.Title(baseRepoName) + "-commit: "
+	ancientSyncCommitSubjectPrefix := fmt.Sprintf("sync(%s/%s)", baseRepoOrg, baseRepoName)
```
sttts: This ancientSyncCommitSubjectPrefix logic can go away. Then we only need the `commitMsgTag`, i.e. `Kubernetes-commit` in the kube case. Makes it much simpler.

artifacts/scripts/util.sh
sttts: Am fine with source or upstream, but we should choose one :)

pkg/git/kube.go
sttts: the flag for `collapsed-kube-commit-mapper` can then be `commit-msg-tag` as well. Makes it much clear why we need that info. Now it is surprising that the tool needs the repo and org.

artifacts/scripts/util.sh
mfojtik: they I would leave it with "source" to make it consistent with the rest of the code

pkg/git/kube.go
mfojtik: well that flag will need to be passed to the scripts as well, right? not sure if we need a new flag :)

sttts: we can construct the commit msg tag in util.sh (maybe we do that already, because we need it for the git-filter-branch call).

artifacts/scripts/util.sh
sttts: :+1:

artifacts/scripts/publish_api.sh
sttts: :+1:

pkg/git/kube.go
mfojtik: that is actually good catch, I haven't replaced Kubernetes-commit in util.sh

mfojtik: @sttts i fixed util.sh to pass the proper commit message via commit_msg_tag... I would go get rid of ancientSyncCommitSubjectPrefix as follow-up, because the number of changes in this PR are getting big, so we should probably start testing this and we can improve later?

PR #16, opened 2018-01-05
Add skipGodeps flag to repo rules

sttts:
TODO:

- [ ] test

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -602,6 +602,10 @@ function git-index-clean() {
 }
 
 function fix-godeps() {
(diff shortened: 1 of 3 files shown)


artifacts/scripts/util.sh
```diff
@@ -596,6 +596,10 @@ function git-index-clean() {
 }
 
 function fix-godeps() {
+    if [ "${SKIP_GODEPS:-}" = true ]; then
```
mfojtik: can we make this more scoped to publisher bot? like: `PUBLISHER_BOT_SKIP_GODEPS`

sttts: done

PR #17, opened 2018-01-08
Externalize the rule definition into a YAML file

mfojtik:
The second commit removes hard-coded Kubernetes rules and move them into a separate YAML file that has to be provided for the publishing-bot via `--rules-file` argument. This argument is required.
This can be either a local filesystem path or an URL form (for easier "fetch from repo directly ;)")...

Other changes:

* Moved the `SkippedSourceBranches` into the rules config
* Removed `./` defaulting for the destination directory

Original rules before this change are available under `cmd/publisher-bot/config/rules_example.yaml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ init-deploy:
 	$(KUBECTL) get StorageClass ssd
 	sed 's/TOKEN/$(shell echo "$(TOKEN)" | base64 | tr -d '\n')/g' artifacts/manifests/secret.yaml | $(KUBECTL) apply -n "$(NAMESPACE)" -f -
 	$(KUBECTL) apply -n "$(NAMESPACE)" -f $(CONFIG)-configmap.yaml
+	$(KUBECTL) apply -n "$(NAMESPACE)" -f $(CONFIG)-rules-configmap.yaml
 	$(KUBECTL) apply -n "$(NAMESPACE)" -f artifacts/manifests/pvc.yaml
 
 run: init-deploy
--- a/cmd/publishing-bot/config/rules.go
+++ b/cmd/publishing-bot/config/rules.go
@@ -16,7 +16,11 @@ type Dependency struct {
 }
 
 func (c Dependency) String() string {
-	return fmt.Sprintf("[repository %s, branch %s, subdir %s]", c.Repository, c.Branch, c.Dir)
+	repo := c.Repository
+	if len(repo) == 0 {
+		repo = "<source>"
+	}
+	return fmt.Sprintf("[repository %s, branch %s, subdir %s]", repo, c.Branch, c.Dir)
 }
 
 type BranchRule struct {
--- a/cmd/publishing-bot/config/rules.go
+++ b/cmd/publishing-bot/config/rules.go
@@ -1,0 +1,60 @@
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Dependency of a piece of code
+type Dependency struct {
+	Repository string `yaml:"repository"`
+	Branch     string `yaml:"branch"`
+	// Dir from repo root
+	Dir string `yaml:"dir,omitempty"`
+}
+
+func (c Dependency) String() string {
+	return fmt.Sprintf("[repository %s, branch %s, subdir %s]", c.Repository, c.Branch, c.Dir)
+}
+
+type BranchRule struct {
+	Name string `yaml:"name"`
+	// k8s.io/* repos the branch rule depends on
+	Dependencies []Dependency `yaml:"dependencies,omitempty"`
+	Source       Dependency   `yaml:"source"`
+}
+
+// a collection of publishing rules for a single destination repo
+type RepositoryRule struct {
+	DestinationRepository string       `yaml:"destination"`
+	Branches              []BranchRule `yaml:"branches"`
+	// if empty (e.g., for client-go), publisher will use its default publish script
+	PublishScript string `yaml:"publish-script"`
+	// not updated when true
+	Skip bool `yaml:"skipped,omitempty"`
+}
+
+type RepositoryRules struct {
+	SkippedSourceBranches []string         `yaml:"skip-source-branches"`
+	Rules                 []RepositoryRule `yaml:"rules"`
+}
+
+// LoadRules loads the repository rules either from the remote HTTP location or
+// a local file path.
+func LoadRules(config *Config) (*RepositoryRules, error) {
+	var (
+		content []byte
+		err     error
+	)
+	content, err = ioutil.ReadFile(config.RulesFile)
+	if err != nil {
+		return nil, err
+	}
+	var rules RepositoryRules
+	if err = yaml.Unmarshal(content, &rules); err != nil {
+		return nil, err
+	}
+	return &rules, nil
+}
--- a/cmd/publishing-bot/main.go
+++ b/cmd/publishing-bot/main.go
@@ -29,6 +29,8 @@ import (
 
 	"time"
 
+	"path/filepath"
+
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
 )
 
@@ -96,6 +98,19 @@ func main() {
 		glog.Fatalf("Target organization cannot be empty")
 	}
 
+	// set the baseRepoPath
+	gopath := os.Getenv("GOPATH")
+	// if the SourceRepo is not kubernetes, use github.com as baseRepoPath
+	baseRepoPath := filepath.Join(gopath, "src", "github.com", cfg.TargetOrg)
+	if cfg.SourceRepo == "kubernetes" {
+		baseRepoPath = filepath.Join(gopath, "src", "k8s.io")
+	}
+
+	// If RULE_FILE_PATH is detected, check if the source repository include rules files.
+	if len(os.Getenv("RULE_FILE_PATH")) > 0 {
+		cfg.RulesFile = filepath.Join(baseRepoPath, cfg.SourceRepo, os.Getenv("RULE_FILE_PATH"))
+	}
+
 	if len(cfg.RulesFile) == 0 {
 		glog.Fatalf("No rules file provided")
 	}
@@ -118,11 +133,7 @@ func main() {
 
 	for {
 		last := time.Now()
-
-		publisher, err := New(&cfg)
-		if err != nil {
-			glog.Fatalf("Failed initialize publisher: %v", err)
-		}
+		publisher := New(&cfg, baseRepoPath)
 
 		if cfg.TokenFile != "" && cfg.GithubIssue != 0 && !cfg.DryRun {
 			// load token
--- a/cmd/publishing-bot/main.go
+++ b/cmd/publishing-bot/main.go
@@ -46,6 +46,7 @@ func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
 	dryRun := flag.Bool("dry-run", false, "do not push anything to github")
 	tokenFile := flag.String("token-file", "", "the file with the github toke")
+	rulesFile := flag.String("rules-file", "", "the file with repository rules")
 	// TODO: make absolute
 	repoName := flag.String("source-repo", "", "the name of the source repository (eg. kubernetes)")
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
@@ -83,6 +84,9 @@ func main() {
 	if *tokenFile != "" {
 		cfg.TokenFile = *tokenFile
 	}
+	if *rulesFile != "" {
+		cfg.RulesFile = *rulesFile
+	}
 
 	if len(cfg.SourceRepo) == 0 || len(cfg.SourceOrg) == 0 {
 		glog.Fatalf("source-org and source-repo cannot be empty")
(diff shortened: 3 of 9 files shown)


cmd/publishing-bot/config/rules_example.yaml
```diff
@@ -0,0 +1,686 @@
+rules:
+- destination-repository: apimachinery
+  source-to-destination:
+  - source:
+      repository: kubernetes
+      branch: master
+      dir: staging/src/k8s.io/apimachinery
+    destination:
+      repository: apimachinery
```
sttts: isn't this the same as the `destination-repository` above?

sttts: Would such a struct do?
```yaml
rules:
- destination: apimachinery
   branches:
     master:
       source: kubernetes
       source-branch: master
       dir: staging/src/k8s.io/apimachinery
       destination-dir: ./  # optional
```

mfojtik: branches can't be map as they have to be ordered, but I tried to simplify this a little.

mfojtik:
@sttts done, ready for review when you get some time :-)

artifacts/scripts/publish_template.sh
```diff
@@ -89,6 +89,7 @@ EXTRA_ARGS=()
 PUSH_SCRIPT=../push-tags-${REPO}-${DST_BRANCH}.sh
 echo "#!/bin/bash" > ${PUSH_SCRIPT}
 chmod +x ${PUSH_SCRIPT}
-/sync-tags --upstream-remote upstream-kube --upstream-branch "${SRC_BRANCH}" \
+/sync-tags --repo-org kubernetes --repo-name kubernetes \
```
mfojtik: @sttts so this needed to be updated to pass the repo-org and repo-name, but I guess this will be different for other repo/orgs and this template can be modified per-usage?

sttts: It's just a constant. We can use a generic one like `upstream`. Just grep in util.sh for it.

cmd/publishing-bot/config/rules_example.yaml
sttts: why do they have to be ordered?

cmd/publishing-bot/main.go
```diff
@@ -46,6 +46,7 @@ func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
 	dryRun := flag.Bool("dry-run", false, "do not push anything to github")
 	tokenFile := flag.String("token-file", "", "the file with the github toke")
+	rulesFile := flag.String("rules-file", "", "the file with repository rules")
```
sttts: How do you imagine this to be used? Should it point to the source repository worktree? How can we make sure that we always have master checked out? We could put a checkout command into the init script in the init container.

mfojtik: @sttts i see two options:

1) Same as config (via configMap), same plumbings
2) Fetch from URL (iow. when you want to keep the rules in the repo). Not sure if fetch from URL is reasonable and we rather should look into the latest cloned sources for the source repo and have some standard path where this file lives.

mfojtik: @sttts i guess for beginning, I would go just with configMap and we can improve based on how we want to manage the rules.

mfojtik:
@sttts this is now rebased on top of the rebranding branch :)

mfojtik:
@sttts rebased

Makefile
```diff
@@ -45,6 +45,7 @@ init-deploy:
 	$(KUBECTL) get StorageClass ssd
 	sed 's/TOKEN/$(shell echo "$(TOKEN)" | base64 | tr -d '\n')/g' artifacts/manifests/secret.yaml | $(KUBECTL) apply -n "$(NAMESPACE)" -f -
 	$(KUBECTL) apply -n "$(NAMESPACE)" -f $(CONFIG)-configmap.yaml
+	$(KUBECTL) apply -n "$(NAMESPACE)" -f $(CONFIG)-rules-configmap.yaml
```
sttts: README.md must be updated to include this file.

cmd/publishing-bot/config/rules.go
```diff
@@ -0,0 +1,85 @@
...
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Coordinate of a piece of code
+type Coordinate struct {
+	Repository string `yaml:"repository"`
+	Branch     string `yaml:"branch"`
+	// Dir from repo root
+	Dir string `yaml:"dir,omitempty"`
```
sttts: do we default this?

cmd/publishing-bot/config/rules.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (c Coordinate) String() string {
+	return fmt.Sprintf("[repository %s, branch %s, subdir %s]", c.Repository, c.Branch, c.Dir)
+}
+
+type BranchRule struct {
+	Source      Coordinate `yaml:"source"`
+	Destination Coordinate `yaml:"destination"`
```
sttts: Please drop destination and just use `Name string` for the branch name.

cmd/publishing-bot/config/rules.go
```diff
@@ -0,0 +1,85 @@
...
+	// k8s.io/* repos the dst dependes on
+	Dependencies []Coordinate `yaml:"dependencies,omitempty"`
```
sttts: replace Coordinate with `type Dependency struct { repository, branch string }`

cmd/publishing-bot/config/rules.go
```diff
@@ -0,0 +1,85 @@
...
+	if err = yaml.Unmarshal(content, &rules); err != nil {
+		return nil, err
+	}
+	return &rules, nil
+}
+
+// fetchFromHTTP fetches the rules from http(s)://
+func fetchFromHTTP(url string) ([]byte, error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
sttts: Don't like InsecureSkipVerify. If we want that, this must be a command line flag.

mfojtik: i think i will remove this for now... we can have init container that will fetch the rules via curl if we need? or just use secret

cmd/publishing-bot/config/rules.go
mfojtik: @sttts the dir is needed for Source (used as a 'subdir')

cmd/publishing-bot/config/rules.go
sttts: Yes, init container can do that. The only advantage: we could fetch the rules on every run (the pod keep running right now, no cronjobs).

mfojtik: @sttts also I think we should have option to get the rules from the repo we fetched :-) like it should have some standard path or an env var like:

REPO_RULES_PATH="./publish-rules.yaml"

sttts: --source-rules-file hack/publish-rules.yaml

cmd/publishing-bot/config/rules.go
mfojtik: yes, we do

Makefile
mfojtik: done.

mfojtik:
@sttts comments addressed, hopefully I haven't screwed the kube rules by refactoring ;-)

cmd/publishing-bot/publisher.go
```diff
@@ -54,6 +54,11 @@ func New(cfg *config.Config) (*PublisherMunger, error) {
 		p.baseRepoPath = filepath.Join(gopath, "src", "k8s.io")
 	}
 
+	// If RULE_FILE_PATH is detected, check if the source repository include rules files.
+	if len(os.Getenv("RULE_FILE_PATH")) > 0 {
```
mfojtik: @sttts with this we can load rules from the source repo... guess we need to reload when we git fetch

cmd/publishing-bot/main.go
```diff
@@ -92,6 +96,10 @@ func main() {
 		glog.Fatalf("Target organization cannot be empty")
 	}
 
+	if len(cfg.RulesFile) == 0 && len(os.Getenv("RULE_FILE_PATH")) == 0 {
```
sttts: the env var reading should also be here. Less magic is better.

sttts: Do we actually need this env var? Can we drop it? We don't have env vars for the other options either.

cmd/publishing-bot/publisher.go
```diff
@@ -71,472 +42,45 @@ type PublisherMunger struct {
...
-
+	p.reposRules = *rules
 	// TODO: re-enable 1.5, 1.6, 1.7 after enforcing every repo branch to update once to pick up new'ish Godep targets
 	// background: when we still had sync commits, we could not correlate upstream commits correctly
-	p.skippedSourceBranches = []string{
-		"release-1.5",
-		"release-1.6",
-	}
-	return p, nil
+	p.skippedSourceBranches = rules.SkippedSourceBranches
```
sttts: comment can be dropped

cmd/publishing-bot/publisher.go
```diff
@@ -71,472 +42,45 @@ type PublisherMunger struct {
...
-				// rule for the metrics release-1.9 branch
-				src:  coordinate{repo: config.SourceRepo, branch: "release-1.9", dir: "staging/src/k8s.io/code-generator"},
-				dst:  coordinate{repo: "code-generator", branch: "release-1.9", dir: "./"},
-				deps: []coordinate{},
-			},
-		},
-		publishScript: "/publish_scripts/publish_code_generator.sh",
+	// NOTE: Order of the repos is sensitive!!!
+	// A dependent repo needs to be published first, so that other repos can vendor its latest revision.
+	rules, err := config.LoadRules(cfg)
```
sttts: just do this call directly for New. Seems to be overkill to have a func that call essentially one func.

cmd/publishing-bot/publisher.go
sttts: do we need PublisherMunger. skippedSourceBranches anymore if we have the rules?

cmd/publishing-bot/publisher.go
```diff
@@ -71,472 +42,45 @@ type PublisherMunger struct {
...
-				dst: coordinate{repo: "sample-controller", branch: "release-1.9", dir: "./"},
-				deps: []coordinate{
-					{repo: "apimachinery", branch: "release-1.9"},
-					{repo: "api", branch: "release-1.9"},
-					{repo: "client-go", branch: "release-6.0"},
-				},
-			},
-		},
-		publishScript: "/publish_scripts/publish_sample_controller.sh",
+	if err := p.readRepositoryRules(cfg); err != nil {
```
sttts: We call this here and in updateSourceRepo. Feels strange.

If we read it from the source repo, we have to update the source repo first before read the rules. Do we have the source repo name without the rules? From the config map?

configs/kubernetes-rules-configmap.yaml
```diff
@@ -0,0 +1,467 @@
...
+          branch: release-1.9
+        - repository: api
+          branch: release-1.9
+        - repository: client-go
+          branch: release-6.0
+      publish-script: /publish_scripts/publish_metrics.sh
+    - destination: code-generator
+      branches:
+      - source:
+          repository: kubernetes
```
sttts: if we have one bot per source repo, we can drop the repository of the source.

sttts:
In short: let's put the source repository and org into the config map, remove both from the rules and read the rules every run after updating the source repo.

cmd/publishing-bot/main.go
mfojtik: @sttts we need the baseRepoPath, but I think we can construct it here

cmd/publishing-bot/publisher.go
mfojtik: probably not

configs/kubernetes-rules-configmap.yaml
mfojtik: it seems like we don't even use this value anywhere

mfojtik:
@sttts code fixed, removed repository: kubernetes and refactored the New() (it is now even more pure)

cmd/publishing-bot/config/rules.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(repo) == 0 {
+		repo = "<source>"
+	}
+	return fmt.Sprintf("[repository %s, branch %s, subdir %s]", repo, c.Branch, c.Dir)
+}
+
+type BranchRule struct {
+	Name string `yaml:"name"`
+	// k8s.io/* repos the branch rule depends on
+	Dependencies []Dependency `yaml:"dependencies,omitempty"`
```
sttts: follow-up: split Dependency and Source. The `Dir` doesn't make sense for the former.

configs/kubernetes-rules-configmap.yaml
```diff
@@ -0,0 +1,425 @@
+kind: ConfigMap
+metadata:
+  name: publisher-rules
+data:
+  config: |
+    skip-source-branches: 
+    - release-1.5
+    - release-1.6
+    - release-1.7
```
sttts: follow-up: re-add 1.7

sttts: https://github.com/kubernetes/publishing-bot/pull/24

cmd/publishing-bot/config/rules.go
sttts: https://github.com/kubernetes/publishing-bot/pull/25

PR #18, opened 2018-01-10
Fix panic with nil github issue number in config

sttts:


cmd/publishing-bot/config/config.go
```diff
@@ -32,5 +32,5 @@ type Config struct {
 	DryRun bool
 
 	// A github issue number to report errors
-	GithubIssue *int `yaml:"github-issue,omitempty"`
+	GithubIssue int `yaml:"github-issue,omitempty"`
```
mfojtik: maybe make this a string and if empty don't post anything? allows more notification delivery tools

sttts: We can change that later when we add another delivery method. Would keep it simple here, just to fix the panic.

mfojtik: fine with that

PR #20, opened 2018-01-12
Remove hardcoded kube

mfojtik:
Reopened https://github.com/kubernetes/publishing-bot/pull/14 seems like GH screwed up or wrong rebase...

his pull attempts to remove all hardcoded kube variables which should allow to use the publisher bot for repositories other than kubernetes/kubernetes.
The defaulting should be preserved, so if SourceRepo is not specified it will default to "kubernetes", thus no changes in how the publisher bot is currently deployed are needed.

### DIFF
--- a/artifacts/scripts/publish_template.sh
+++ b/artifacts/scripts/publish_template.sh
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 6 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos kubernetes_remote is_library"
+if [ ! $# -eq 9 ]; then
+    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos kubernetes_remote subdirectory source_repo_org source_repo_name is_library"
     exit 1
 fi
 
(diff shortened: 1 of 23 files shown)


configs/kubernetes-nightly-configmap.yaml
```diff
@@ -3,6 +3,8 @@ metadata:
   name: publisher-config
 data:
   config: |
+    repo-org: kubernetes-nightly
```
mfojtik: @sttts check?

mfojtik:
this is now tested and should work fine

artifacts/scripts/publish_template.sh
```diff
@@ -37,7 +37,7 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 6 ]; then
+if [ ! $# -eq 9 ]; then
     echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos kubernetes_remote is_library"
```
sttts: error text is outdated

cmd/publishing-bot/config/config.go
```diff
@@ -21,9 +21,11 @@ type Config struct {
 	// the organization to publish into, e.g. k8s-publishing-bot or kubernetes-nightly
 	TargetOrg string `yaml:"target-org"`
 
-	// the source repo, e.g. "kubernetes"
-	// TODO: make this absolute
-	SourceRepo string `yaml:"source-repo,omitempty"`
+	// the source repo name, e.g. "kubernetes"
+	SourceRepoName string `yaml:"repo-name"`
```
sttts: source-org and source-repo would be clearer IMO. Also in flags.

configs/example-configmap.yaml
```diff
@@ -3,6 +3,9 @@ metadata:
   name: publisher-config
 data:
   config: |
+    # this specifies the source repository coordinates (org/name)
+    repo-org: <your-github-org-or-user>
```
sttts: `<upstream-org-or-user>`, it's usually not "your org"

configs/kubernetes-configmap.yaml
```diff
@@ -3,6 +3,8 @@ metadata:
   name: publisher-config
 data:
   config: |
+    repo-org: kubernetes
```
sttts: as written above, I don't like the `repo-` prefix. Why not `source-org` and `source-repo`? Matches `target-org` below.

mfojtik:
@sttts all comments addressed, I also renamed the fields and flags to make it all consistent.

artifacts/scripts/publish_template.sh
```diff
@@ -82,13 +88,14 @@ fi
...
+sync_repo "${SOURCE_REPO_ORG}" "${SOURCE_REPO_NAME}" "${SUBDIR}" "${SRC_BRANCH}" "${DST_BRANCH}" "${SOURCE_REMOTE}" "${DEPS}" "${IS_LIBRARY}"
 
 # add tags
 EXTRA_ARGS=()
 PUSH_SCRIPT=../push-tags-${REPO}-${DST_BRANCH}.sh
 echo "#!/bin/bash" > ${PUSH_SCRIPT}
 chmod +x ${PUSH_SCRIPT}
-/sync-tags --upstream-remote upstream-kube --upstream-branch "${SRC_BRANCH}" \
+/sync-tags --source-org ${SOURCE_REPO_ORG} --source-repo ${SOURCE_REPO_NAME} \
+           --upstream-remote upstream --upstream-branch "${SRC_BRANCH}" \
```
sttts: follow-up: rename --upstream-remote -> --source-remote
follow-up: rename --upstream-branch -> --source-branch

PR #21, opened 2018-01-15
Return github errors, don't call glog.Fatalf

sttts:


sttts:
@mfojtik ptal. This is becoming critical because the bot spams the github issue every 8 minutes on error

PR #22, opened 2018-01-15
Rename upstream->source in flags and pass commit-msg-tag directly

sttts:


cmd/collapsed-kube-commit-mapper/main.go
```diff
@@ -105,17 +104,17 @@ func main() {
 	}
 
 	// get first-parent commit list of upstream branch
-	kUpstreamBranch, err := r.ResolveRevision(plumbing.Revision(*upstreamBranch))
+	kUpstreamBranch, err := r.ResolveRevision(plumbing.Revision(*sourceBranch))
```
mfojtik: at some point we should get rid of k* prefix ;-)

sttts: indeed.

PR #23, opened 2018-01-17
Add init-repo command

mfojtik:
This will replace the init_repositories.sh script that only works for kube staging dirs.
The init-repo command will read the config and external rules. Based on rules it will determine what fork repositories it should clone. 

(only last commit is interesting, the first two are part of different PR)


mfojtik:
@sttts last piece of puzzle :-)

mfojtik:
FYI: this is currently untested, will give it a run tomorrow morning.

cmd/init-repo/main.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/publishing-bot/cmd/publishing-bot/config"
+)
+
+const (
+	depCommit   = "7c44971bbb9f0ed87db40b601f2d9fe4dffb750d"
+	godepCommit = "tags/v79"
+)
+
+var (
+	SystemGoPath = os.Getenv("GOPATH")
```
mfojtik: i'm 90% sure there is some special package in Go that has this (like os.GOPATH or something ;-) but googling it is hell.

mfojtik:
@sttts i tested the `init-repo` it works like a charm (also for fake_openshift/origin ;-)

mfojtik:
@sttts https://gist.github.com/mfojtik/a6420ec43d8430cddec4d1f1ff04c754 this is the farest I got.

It seems like the scripts are working fine, the `Origin-commit:` is used properly, but we are missing the initial commit in the fork repo I guess. 

The rules used: https://gist.github.com/mfojtik/ed6bf3fa9c2775e91b6ebfbafb431815

cmd/publishing-bot/publisher.go
```diff
@@ -155,6 +155,19 @@ func (p *PublisherMunger) construct() error {
 			}
 			return strings.Join(depStrings, ",")
 		}
+		if len(repoRules.PublishScript) == 0 {
+			// TODO: Move this to main.go?
```
sttts: yes, please

cmd/publishing-bot/publisher.go
```diff
@@ -155,6 +155,19 @@ func (p *PublisherMunger) construct() error {
...
+			basePublishScriptPath := "./publish_scripts"
+			if len(os.Getenv("BASE_PUBLISH_SCRIPT_PATH")) > 0 {
+				basePublishScriptPath = os.Getenv("BASE_PUBLISH_SCRIPT_PATH")
+			}
+			repoPublishScriptPath := filepath.Join(basePublishScriptPath, "publish_"+repoRules.DestinationRepository+".sh")
+			if _, err := os.Stat(repoPublishScriptPath); err == nil {
+				repoRules.PublishScript = repoPublishScriptPath
```
sttts: follow-up: get rid of these scripts and put the smoke tests into the rules

sttts:
Some nits

cmd/publishing-bot/publisher.go
mfojtik: actually can't be done because this depends on the rules to be loaded.

cmd/publishing-bot/publisher.go
mfojtik: https://github.com/kubernetes/publishing-bot/issues/26

mfojtik:
@sttts determining the location of publishingScript in repo depends on the repo be available locally and depends on the rules to be already loaded... So I don't think we can move that to main.go for now.

We can refactor later maybe.

mfojtik:
@sttts k, add `BasePublishScriptsPath` into Config.

PR #28, opened 2018-01-23
add /run endpoint that force the bot to run immediately

mfojtik:


mfojtik:
@kargakis i'm planning to add prow bot that will receive pushes to a repo and then hit this endpoint. (micro-service all the things!)

cmd/publishing-bot/healthz.go
```diff
@@ -29,7 +29,8 @@ import (
 )
 
 type Healthz struct {
```
sttts: rename the struct and file, maybe "server".

cmd/publishing-bot/healthz.go
```diff
@@ -77,7 +79,18 @@ func (h *Healthz) Run(port int) error {
 	return nil
 }
 
-func (h *Healthz) handler(w http.ResponseWriter, r *http.Request) {
+func (h *Healthz) runHandler(w http.ResponseWriter, r *http.Request) {
+	if h.RunChan == nil {
+		http.Error(w, "run channel is closed", http.StatusInternalServerError)
+		return
+	}
+	h.mutex.RLock()
```
kargakis: Why do you need a mutex to protect a channel?

cmd/publishing-bot/main.go
```diff
@@ -119,10 +119,14 @@ func main() {
 		glog.Fatalf("No rules file provided")
 	}
 
+	runChan := make(chan bool, 1)
+	defer close(runChan)
```
sttts: wouldn't close it. might lead to a panic from the server on shutdown.

cmd/publishing-bot/healthz.go
mfojtik: leftover

cmd/publishing-bot/healthz.go
```diff
@@ -77,7 +79,18 @@ func (h *Healthz) Run(port int) error {
...
+	h.RunChan <- true
```
sttts: No need for the mutex here. We don't change the channel ever.

sttts: And do:
```
select {
  case h.RunChan <- true:
  default:
}
```

cmd/publishing-bot/main.go
kargakis: Properly handle SIGINT and SIGTERM if you want graceful termination.

mfojtik: we probably don't care about graceful termination, but would be interesting what happen to a running bot when something kill the pod in a middle of the sync.

cmd/publishing-bot/healthz.go
mfojtik: @sttts do where?

mfojtik:
@sttts fixed up

cmd/publishing-bot/main.go
kargakis: Yeah, I think you do care. We had issues in prow because we werent gracefully terminating and bots would be killed in the middle of an update, leaving partial Github updates behind.

mfojtik:
added non-blocking write, tested and it works perfectly fine

mfojtik:
@sttts i'll make an issue to handle unexpected shutdowns

PR #32, opened 2018-01-25
do not fail when Godeps/Godeps.json is removed

mfojtik:
start from a given Kubernetes-commit and do not try to move to a merge
commit.

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -214,12 +214,12 @@ sync_repo() {
                 echo "Couldn't find a ${commit_msg_tag} commit SHA in any commit on ${dst_branch}."
                 return 1
             fi
-            local k_base_merge=$(git-find-merge ${k_base_commit} upstream/${src_branch})
-            if [ -z "${k_base_merge}" ]; then
-                echo "Didn't find merge commit of source commit ${k_base_commit}. Odd."
-                return 1
-            fi
-            git branch -f filtered-branch-base ${k_base_merge} >/dev/null
+            #local k_base_merge=$(git-find-merge ${k_base_commit} upstream/${src_branch})
+            #if [ -z "${k_base_merge}" ]; then
+            #    echo "Didn't find merge commit of source commit ${k_base_commit}. Odd."
+            #    return 1
+            #fi
+            git branch -f filtered-branch-base ${k_base_commit} >/dev/null
 
             echo "Rewriting upstream branch ${src_branch} to only include commits for ${subdirectory}."
             filter-branch "${commit_msg_tag}" "${subdirectory}" filtered-branch filtered-branch-base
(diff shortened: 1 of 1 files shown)


artifacts/scripts/util.sh
```diff
@@ -214,12 +214,12 @@ sync_repo() {
                 echo "Couldn't find a ${commit_msg_tag} commit SHA in any commit on ${dst_branch}."
                 return 1
             fi
-            local k_base_merge=$(git-find-merge ${k_base_commit} upstream/${src_branch})
-            if [ -z "${k_base_merge}" ]; then
-                echo "Didn't find merge commit of source commit ${k_base_commit}. Odd."
-                return 1
-            fi
-            git branch -f filtered-branch-base ${k_base_merge} >/dev/null
+            #local k_base_merge=$(git-find-merge ${k_base_commit} upstream/${src_branch})
```
sttts: please remove the code

artifacts/scripts/util.sh
```diff
@@ -499,7 +499,10 @@ EOF
 # amend-godeps-at checks out the Godeps.json at the given commit and amend it to the previous commit.
 function amend-godeps-at() {
     if [ -f Godeps/Godeps.json ]; then
+        set +e
         git checkout ${f_mainline_commit} Godeps/Godeps.json # reset to mainline state which is guaranteed to be correct
```
sttts: I still wonder how the Godeps.json can exist here, but is gone in `${f_mainline_commit}`.

sttts: We should add a skipGodeps check in amend-godeps-at and also where godep-changes is called (or even in that func).

artifacts/scripts/util.sh
```diff
@@ -214,12 +214,12 @@ sync_repo() {
...
-                echo "Didn't find merge commit of source commit ${k_base_commit}. Odd."
-                return 1
-            fi
-            git branch -f filtered-branch-base ${k_base_merge} >/dev/null
+            #local k_base_merge=$(git-find-merge ${k_base_commit} upstream/${src_branch})
+            #if [ -z "${k_base_merge}" ]; then
+            #    echo "Didn't find merge commit of source commit ${k_base_commit}. Odd."
+            #    return 1
+            #fi
+            git branch -f filtered-branch-base ${k_base_commit} >/dev/null
```
sttts: at this position we can check with `git diff --exit-code --quiet ${filtered-branch-base} ${filtered-branch} -- "${subdirectory}"` whether there was any change. If there hasn't been one, we can return and safe the runtime of the filter-branch rewrite. This will easily cut down the runtime by 30sec per branch.

artifacts/scripts/util.sh
mfojtik: that will require plumbing all the way down from sync-repo, not sure if it is worth for this case.

artifacts/scripts/util.sh
mfojtik: not sure I understand... you want to replace git branch -f with git diff?

sttts: https://github.com/kubernetes/publishing-bot/pull/41

Issue #26, opened 2018-01-22
Get rid of these scripts and put the smoke tests into the rules


mfojtik:
Now that the rules are external, we should move the publish script into rules (YAML) and get rid of specific scripts in artifacts/scripts.

sttts:
Yes, please!!! Only the smoke tests though, most of the scripts is boilerplate.

PR #34, opened 2018-02-10
Switch publish script into smoke tests in rules yaml

sttts:
Fixes https://github.com/kubernetes/publishing-bot/issues/26


cmd/publishing-bot/main.go
```diff
@@ -53,7 +53,7 @@ func main() {
 	repoName := flag.String("source-repo", "", "the name of the source repository (eg. kubernetes)")
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	targetOrg := flag.String("target-org", "", `the target organization to publish into (e.g. "k8s-publishing-bot")`)
-	basePublishScriptPath := flag.String("base-publish-script-path", "./publish_scripts", `the base path in source repo where bot will look for publishing scripts`)
+	basePublishScriptPath := flag.String("base-publish-script-path", "/publish_scripts", `the base path in source repo where bot will look for publishing scripts`)
```
mfojtik: i think this was meant to be a relative path inside the repo

cmd/publishing-bot/config/rules.go
```diff
@@ -48,8 +48,8 @@ type BranchRule struct {
 type RepositoryRule struct {
 	DestinationRepository string       `yaml:"destination"`
 	Branches              []BranchRule `yaml:"branches"`
-	// if empty (e.g., for client-go), publisher will use its default publish script
-	PublishScript string `yaml:"publish-script"`
+	SmokeTest             string       `yaml:"smoke-test"` // a multiline bash script
+	Library               bool         `yaml:"library"`
```
mfojtik: omitempty?

mfojtik:
one nit, otherwise LGTM

cmd/publishing-bot/main.go
sttts: It's relative to the current working dir. I think before this PR it was coincidently the root of the container. Will check. I think we have to restore the current working dir correctly. Then the `./` is fine.

cmd/publishing-bot/config/rules.go
sttts: well, default is false. But printing it (if we ever print rules at all ;-) ) makes sense.

mfojtik: generating (serializing) is ugly when we default empty fields ;-)

cmd/publishing-bot/main.go
sttts: fixed

cmd/publishing-bot/config/rules.go
sttts: fixed

PR #35, opened 2018-02-10
Add required-packages field to include code-generator

sttts:
Fixes https://github.com/kubernetes/sample-controller/issues/6.

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 9 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos kubernetes_remote subdirectory source_repo_org source_repo_name is_library"
+if [ ! $# -eq 10 ]; then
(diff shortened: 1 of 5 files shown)


artifacts/scripts/construct.sh
```diff
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 9 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos kubernetes_remote subdirectory source_repo_org source_repo_name is_library"
+if [ ! $# -eq 10 ]; then
+    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_k8s.io_packages kubernetes_remote subdirectory source_repo_org source_repo_name is_library"
```
mfojtik: is this k8s specific?

mfojtik:
LGTM

artifacts/scripts/construct.sh
sttts: no, not at all actually. Bad variable name. Will fix.

sttts: Fixed

kargakis:
cool

sttts:
For reference, this won't be as easy with dep: https://github.com/golang/dep/issues/1696

We can only copy in the files without caring about dep being unhappy about the mutated vendor/ dir.

PR #40, opened 2018-02-13
allow to fetch rules from URL

mfojtik:
Detects the URL in `-rule-file` value and fetch the rules from remote URL.


sttts:
travis is unhappy

mfojtik:
@sttts should be happy now

PR #41, opened 2018-02-13
Skip all slow steps if there are no commits against the subdirectory

sttts:


nikhita:
/remove-lifecycle stale

this looks ready to be merged

sttts:
/hold

I question now that this PR works at all: even if there is no change in a subdirectory (e.g. staging/src/k8s.io/client-go), the dependencies might have changes (e.g. staging/src/k8s.io/apimachinery). And then we have to update at least the Godeps to avoid inconsistent Godep.json files in different sub-projects.

nikhita:
@sttts Should this be closed?

sttts:
With the recent speed improvement, I don't think we need this.

PR #44, opened 2018-02-16
fix rule loading

mfojtik:
@sttts critical fix :( Without this it will always fallback to downloading from URL which will result in 0 rules.


sttts:
And I was happy that the bot runs again:
```
I0216 11:04:22.734465       1 publisher_logger.go:69] git fetch origin
I0216 11:04:26.219915       1 publisher.go:73] Loaded 0 repository rules from /etc/publisher-rules/config
```
😁

mfojtik:
hehe :-)

Issue #49, opened 2018-02-28
Fix or update build files when syncing


jayconrod:
A build failure was originally reported in bazelbuild/rules_go#1356. It appears that Bazel BUILD files in `api`, `apimachinery`, and `client-go` refer to non-existent packages.

### Steps to reproduce

* In an empty workspace, create an empty BUILD file and a WORKSPACE file like below:

```
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "4b14d8dd31c6dbaf3ff871adcd03f28c3274e42abc855cb8fb4d01233c0154dc",
    url = "https://github.com/bazelbuild/rules_go/releases/download/0.10.1/rules_go-0.10.1.tar.gz",
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "6228d9618ab9536892aa69082c063207c91e777e51bd3c5544c9c060cafe1bd8",
    url = "https://github.com/bazelbuild/bazel-gazelle/releases/download/0.10.0/bazel-gazelle-0.10.0.tar.gz",
)

load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")

go_rules_dependencies()

go_register_toolchains()

load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")

gazelle_dependencies()

load("@bazel_gazelle//:def.bzl", "go_repository")

git_repository(
    name = "io_k8s_api",
    remote = "https://github.com/kubernetes/api",
    tag = "kubernetes-1.9.3",
)

git_repository(
    name = "io_k8s_apimachinery",
    remote = "https://github.com/kubernetes/apimachinery",
    tag = "kubernetes-1.9.3",
)

git_repository(
    name = "io_k8s_client_go",
    remote = "https://github.com/kubernetes/client-go",
    tag = "v6.0.0",
)
```

* Run this command:

```
$ bazel build @io_k8s_client_go//kubernetes/typed/core/v1:go_default_library
ERROR: /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/533bcca7b8c77451fbaf7e2a9f7692b3/external/io_k8s_client_go/kubernetes/typed/core/v1/BUILD:8:1: no such package '@io_k8s_client_go//vendor/k8s.io/apimachinery/pkg/types': BUILD file not found on package path and referenced by '@io_k8s_client_go//kubernetes/typed/core/v1:go_default_library'.
[...]

PR #50, opened 2018-02-28
Implement RecursiveDeletePatterns and delete BUILD files in kube rules

sttts:
Fixes https://github.com/kubernetes/publishing-bot/issues/49
Fixes https://github.com/kubernetes/client-go/issues/189

- [x] test this with a full run against kube

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -133,6 +133,11 @@ sync_repo() {
     local deps="${7:-""}"
     local required_packages="${8:-""}"
     local is_library="${9}"
+
+    shift 9
+
+    local recursive_delete_pattern="${1}"
+
     local commit_msg_tag="${source_repo_name^}-commit"
     readonly subdirectory src_branch dst_branch kubernetes_remote deps is_library
 
@@ -168,7 +173,7 @@ sync_repo() {
     local f_mainline_commits=""
     if [ "${new_branch}" = "true" ] && [ "${src_branch}" = master ]; then
         # new master branch
-        filter-branch "${commit_msg_tag}" "${subdirectory}" ${src_branch} filtered-branch
+        filter-branch "${commit_msg_tag}" "${subdirectory}" "${recursive_delete_pattern}" ${src_branch} filtered-branch
 
         # find commits on the main line (will mostly be merges, but could be non-merges if filter-branch dropped
         # the corresponding fast-forward merge and left the feature branch commits)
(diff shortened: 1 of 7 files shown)


artifacts/scripts/util.sh
```diff
@@ -533,8 +538,13 @@ function commit-subject() {
...
     local commit_msg_tag="${1}"
     local subdirectory="${2}"
+    local recursive_delete_pattern="${3}"
     echo "Running git filter-branch ..."
-    git filter-branch -f --msg-filter 'awk 1 && echo && echo "'"${commit_msg_tag}"': ${GIT_COMMIT}"' --subdirectory-filter "${subdirectory}" -- ${3} ${4} >/dev/null
+    local index_filter=""
+    if [ -n "${recursive_delete_pattern}" ]; then
+        index_filter="git rm --cached --ignore-unmatch -r ${recursive_delete_pattern}"
+    fi
+    git filter-branch -f --index-filter "${index_filter}" --msg-filter 'awk 1 && echo && echo "'"${commit_msg_tag}"': ${GIT_COMMIT}"' --subdirectory-filter "${subdirectory}" -- ${4} ${5} >/dev/null
```
sttts: we will see how much this slows down the publishing run

sttts:
Tested a full run successfully. Here is the client-go BUILD removal: https://github.com/sttts/client-go/commit/d685b9c9f6e502a793237d3b9e0cb39529483f5f

configs/kubernetes-rules-configmap.yaml
```diff
@@ -6,6 +6,9 @@ data:
     skip-source-branches:
     - release-1.5
     - release-1.6
+    recursive-delete-patterns:
+    - BUILD
```
ixdy: might want to also catch `BUILD.bazel` files.

sttts: That's only at the kubernetes root, isn't it?

sttts: I mean: not in staging/src/k8s.io/*

ixdy: right, we generally use `BUILD` in the kubernetes repo, though apparently `BUILD.bazel` is now preferred. I vaguely intend to switch at some point, but if/when I do that I guess I can also update the bot at the same time.

sttts: Then let's add it. It does not hurt.

sttts: done

sttts:
@ixdy @caesarxuchao this needs a lgtm to deploy it for 1.10.

artifacts/scripts/util.sh
```diff
@@ -243,6 +248,19 @@ sync_repo() {
...
+    # remove existing recursive-delete-pattern files. After a first removal commit, the filter-branch command
+    # will filter them out from upstream commits.
+    if [ -n "${recursive_delete_pattern}" ]; then
+        local split_recursive_delete_pattern
+        read -r -a split_recursive_delete_pattern <<< "${recursive_delete_pattern}"
+        git rm -q --ignore-unmatch -r "${split_recursive_delete_pattern[@]}"
+        git add -u
+        if ! git-index-clean; then
+            echo "Deleting files recursively: ${recursive_delete_pattern}"
+            git commit -q -m "sync: initially remove files ${recursive_delete_pattern}"
```
mfojtik: can the commit list the files it deleted? (instead of pattern) (or list the files being deleted in log).

mfojtik:
one comment, LGTM otherwise

artifacts/scripts/util.sh
sttts: That's one file per directory, and for each commit. That's spam the logs a lot.

sttts: actually, here we can be verbose. It's just the initial commit. Done.

PR #53, opened 2018-03-06
WIP: add support for rules-file-branches argument

mfojtik:
This add support for `--rules-file-branches` argument that the bot will use to read the rules file from. This allow to have separate rule file per branch.

### DIFF
--- a/cmd/init-repo/main.go
+++ b/cmd/init-repo/main.go
@@ -13,6 +13,8 @@ import (
 
 	"strings"
 
+	"net/url"
+
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
 )
 
@@ -40,6 +42,7 @@ func main() {
 	repoName := flag.String("source-repo", "", "the name of the source repository (eg. kubernetes)")
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	rulesFile := flag.String("rules-file", "", "the file with repository rules")
+	rulesFileBranches := flag.String("rules-file-branches", "master", "list of source repository branches we look for the rules")
 	targetOrg := flag.String("target-org", "", `the target organization to publish into (e.g. "k8s-publishing-bot")`)
 	skipGodep := flag.Bool("skip-godep", false, `skip godeps installation and godeps-restore`)
 	skipDep := flag.Bool("skip-dep", false, `skip 'dep'' installation`)
@@ -71,7 +74,9 @@ func main() {
 	if cfg.SourceRepo != "kubernetes" {
 		BaseRepoPath = filepath.Join(SystemGoPath, "src", "github.com", cfg.TargetOrg)
 	}
-
+	if *rulesFileBranches != "" {
+		cfg.RuleSourceBranches = *rulesFileBranches
+	}
 	if *rulesFile != "" {
 		cfg.RulesFile = *rulesFile
 	}
@@ -84,17 +89,8 @@ func main() {
 		glog.Fatalf("Target organization cannot be empty")
 	}
 
-	// If RULE_FILE_PATH is detected, check if the source repository include rules files.
-	if len(os.Getenv("RULE_FILE_PATH")) > 0 {
-		cfg.RulesFile = filepath.Join(BaseRepoPath, cfg.SourceRepo, os.Getenv("RULE_FILE_PATH"))
-	}
-
 	if len(cfg.RulesFile) == 0 {
 		glog.Fatalf("No rules file provided")
-	}
-	rules, err := config.LoadRules(cfg.RulesFile)
-	if err != nil {
-		glog.Fatalf("Failed to load rules: %v", err)
 	}
 
 	if err := os.MkdirAll(BaseRepoPath, os.ModePerm); err != nil {
@@ -107,10 +103,40 @@ func main() {
 	if !*skipDep {
 		installDep()
 	}
-
 	cloneSourceRepo(cfg, *skipGodep)
-	for _, rule := range rules.Rules {
-		cloneForkRepo(cfg, rule.DestinationRepository)
+
+	var (
+		rules *config.RepositoryRules
+		err   error
+	)
+
+	branches := strings.Split(cfg.RuleSourceBranches, ",")
+	clonedRepos := map[string]bool{}
+
+	for _, sourceBranch := range branches {
+		if strings.HasPrefix(cfg.RulesFile, "/") {
+			// The rules are specified as absolute file path
+			rules, err = config.LoadRulesFromFile(cfg.RulesFile)
+		} else {
+			// The rules are specified as URL
+			if ruleUrl, err := url.ParseRequestURI(cfg.RulesFile); err == nil && len(ruleUrl.Host) > 0 {
+				rules, err = config.LoadRulesFromURL(ruleUrl)
+			} else {
+				// The rules are specified as relative path in repository dir
+				rules, err = config.LoadRulesFromRepo(cfg.RulesFile, filepath.Join(BaseRepoPath, cfg.SourceRepo), sourceBranch)
+			}
+		}
+		if err != nil || rules == nil {
+			glog.Fatalf("Failed to load rules %q: %v", cfg.RulesFile, err)
+		}
+		for _, rule := range rules.Rules {
+			if _, exists := clonedRepos[rule.DestinationRepository]; exists {
+				glog.Infof("Already cloned fork repository %s, skipping ...", rule.DestinationRepository)
+				continue
+			}
+			cloneForkRepo(cfg, rule.DestinationRepository)
+			clonedRepos[rule.DestinationRepository] = true
+		}
 	}
 }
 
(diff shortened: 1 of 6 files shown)


cmd/publishing-bot/config/config.go
```diff
@@ -27,6 +27,10 @@ type Config struct {
 	// the source repo org name, e.g. "kubernetes"
 	SourceOrg string `yaml:"source-org"`
 
+	// list of source branches to run the bot from
+	// if specified, the bot will checkout the given branch before reading the rule file
+	SourceBranches string `yaml:"source-branches"`
```
sttts: Maybe RuleSourceBranches?

cmd/publishing-bot/config/rules.go
```diff
@@ -68,15 +70,21 @@ type RepositoryRules struct {
...
 		content []byte
 		err     error
 	)
+
 	if ruleUrl, err := url.ParseRequestURI(ruleFile); err == nil && len(ruleUrl.Host) > 0 {
+		// The sourceBranch will be ignored when reading from URL
 		content, err = readFromUrl(ruleUrl)
 	} else {
-		content, err = ioutil.ReadFile(ruleFile)
+		if len(sourceBranch) > 0 {
```
sttts: can we get rid of this if clause by defaulting the sourceBranch to "master" outside of this func? Actually, I would split this function up. We model a sum type in the input by multiple parameters. That's ugly.

cmd/publishing-bot/publisher.go
```diff
@@ -65,7 +65,7 @@ func (p *PublisherMunger) updateSourceRepo() (string, error) {
 		return "", fmt.Errorf("failed running %v on %q repo: %v", strings.Join(cmd.Args, " "), p.config.SourceRepo, err)
 	}
 
-	rules, err := config.LoadRules(p.config.RulesFile)
+	rules, err := config.LoadRules(p.config.RulesFile, sourceBranch, repoDir)
```
sttts: can we have the checkout here and remove the branch handling for LoadRules?

cmd/publishing-bot/publisher.go
```diff
@@ -244,25 +244,57 @@ func (p *PublisherMunger) publish() error {
...
 
-	hash, err := p.updateSourceRepo()
-	if err != nil {
-		p.plog.Errorf("%v", err)
-		p.plog.Flush()
-		return p.plog.Logs(), hash, err
+	// If --branches is specified, re-run the bot with different set of rules
+	// per each specified branch.
+	sourceBranches := strings.Split(p.config.SourceBranches, ",")
+	if len(sourceBranches) == 0 {
```
sttts: `len(sourceBranches) == 0` should mean that nothing is done. Defaulting should be outside of the func.

cmd/publishing-bot/publisher.go
mfojtik: i dunno what consequences the checkout can have here (i guess we will then checkout to rule branches anyway?)

cmd/init-repo/main.go
```diff
@@ -84,18 +89,9 @@ func main() {
...
 
-	// If RULE_FILE_PATH is detected, check if the source repository include rules files.
-	if len(os.Getenv("RULE_FILE_PATH")) > 0 {
-		cfg.RulesFile = filepath.Join(BaseRepoPath, cfg.SourceRepo, os.Getenv("RULE_FILE_PATH"))
-	}
-
 	if len(cfg.RulesFile) == 0 {
 		glog.Fatalf("No rules file provided")
 	}
-	rules, err := config.LoadRules(cfg.RulesFile)
```
mfojtik: @sttts I broke this into 3 separate functions each having its own logic to handle the sources...

With this PR we will handle reading of the rule file using:

```
--rules-file=/etc/foo/rules.yaml              # Absolute path
--rules-file=configs/rules.yaml               # Relative to source repository root (rules-file-branches affecting the branch)
--rules-file=https://foo.com/rules.yaml  # URL

```

cmd/init-repo/main.go
```diff
@@ -13,6 +13,8 @@ import (
 
 	"strings"
```
sttts: nit: no new line. Goland still cannot do this properly.

cmd/init-repo/main.go
```diff
@@ -107,10 +103,40 @@ func main() {
...
 		installDep()
 	}
-
 	cloneSourceRepo(cfg, *skipGodep)
-	for _, rule := range rules.Rules {
-		cloneForkRepo(cfg, rule.DestinationRepository)
+
+	var (
+		rules *config.RepositoryRules
+		err   error
```
sttts: why a global err?

cmd/init-repo/main.go
```diff
@@ -107,10 +103,40 @@ func main() {
 	if !*skipDep {
 		installDep()
 	}
-
 	cloneSourceRepo(cfg, *skipGodep)
-	for _, rule := range rules.Rules {
-		cloneForkRepo(cfg, rule.DestinationRepository)
+
+	var (
+		rules *config.RepositoryRules
```
sttts: why global rules?

cmd/init-repo/main.go
```diff
@@ -107,10 +103,40 @@ func main() {
...
+	)
+
+	branches := strings.Split(cfg.RuleSourceBranches, ",")
+	clonedRepos := map[string]bool{}
+
+	for _, sourceBranch := range branches {
+		if strings.HasPrefix(cfg.RulesFile, "/") {
+			// The rules are specified as absolute file path
+			rules, err = config.LoadRulesFromFile(cfg.RulesFile)
+		} else {
```
sttts: `else if`. The nesting is not needed, is it?

cmd/init-repo/main.go
```diff
@@ -107,10 +103,40 @@ func main() {
...
+			// The rules are specified as URL
+			if ruleUrl, err := url.ParseRequestURI(cfg.RulesFile); err == nil && len(ruleUrl.Host) > 0 {
+				rules, err = config.LoadRulesFromURL(ruleUrl)
+			} else {
+				// The rules are specified as relative path in repository dir
+				rules, err = config.LoadRulesFromRepo(cfg.RulesFile, filepath.Join(BaseRepoPath, cfg.SourceRepo), sourceBranch)
+			}
+		}
+		if err != nil || rules == nil {
```
sttts: can rules be nil and `err == nil` at the same time?

cmd/init-repo/main.go
```diff
@@ -40,6 +42,7 @@ func main() {
 	repoName := flag.String("source-repo", "", "the name of the source repository (eg. kubernetes)")
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	rulesFile := flag.String("rules-file", "", "the file with repository rules")
+	rulesFileBranches := flag.String("rules-file-branches", "master", "list of source repository branches we look for the rules")
```
sttts: rules-file-branches vs. rule-source-branches. We should unify them.

cmd/publishing-bot/config/config.go
```diff
@@ -33,6 +33,9 @@ type Config struct {
 	// the file that contain the repository rules
 	RulesFile string `yaml:"rules-file"`
```
sttts: godoc about absolute, relative and url

cmd/publishing-bot/config/rules.go
```diff
@@ -9,6 +9,8 @@ import (
 	"net/url"
 	"time"
```
sttts: nit: no line

cmd/publishing-bot/main.go
```diff
@@ -90,6 +91,9 @@ func main() {
 	if *rulesFile != "" {
 		cfg.RulesFile = *rulesFile
 	}
+	if *rulesFileBranches != "" {
```
sttts: isn't this "master" by default, i.e. the config is always overridden?

cmd/publishing-bot/publisher.go
```diff
@@ -65,10 +67,27 @@ func (p *PublisherMunger) updateSourceRepo() (string, error) {
...
+		}
+	}
 	if err != nil {
 		return "", err
 	}
+	if rules == nil {
+		return "", fmt.Errorf("Unable to load rule file from %q", p.config.RulesFile)
+	}
+
 	p.reposRules = *rules
```
sttts: why do we read the rules here and not in the main loop? Overriding the rules in `p` is ugly, now that they actually differ per branch.

cmd/publishing-bot/publisher.go
```diff
@@ -244,25 +263,48 @@ func (p *PublisherMunger) publish() error {
...
+		hashes[branch] = ""
 	}
-	if err := p.construct(); err != nil {
-		p.plog.Errorf("%v", err)
-		p.plog.Flush()
-		return p.plog.Logs(), hash, err
+
+	var errors []error
+	for _, sourceBranch := range sourceBranches {
+		hash, err := p.updateSourceRepo(sourceBranch)
```
sttts: Am not convinced to do a "git fetch" here inside the loop. This could lead to inconsistencies if the upstream repo changes in between.

PR #60, opened 2018-04-06
Indent godep output with tabs

sttts:
Now we get ugly re-indention commit in certain situation. Godeps.json should always be indented with tabs.


cblecker:
/lgtm
👍

PR #61, opened 2018-04-09
Squash godep commits or skip all of them if nothing changed

sttts:
This stops the commit loop visible in https://github.com/kubernetes/apiextensions-apiserver/commits/master.

We had a number of steps, each avoiding commits if nothing changed, but without a check for all of them in a row (leading to the loop):
- remove bazel BUILD files
- run godep save/restore
- add expicitly required packages (like k8s.io/code-generator).

Each of these alone creates a diff, but all 3 squashed are empty.


sttts:
/cc @liggitt 

This was the commit loop you noticed.

PR #62, opened 2018-04-09
Keep 7 days full log history in (rotated) run.log

sttts:
Logs are in `<repoBasePath>/run.log`, rotated on each run, kept for 7 days.


nikhita:
This is awesome.

/lgtm

PR #69, opened 2018-06-29
Replace non-existing clone flag --no-tags in git 2.1.4 with bash tag removal

sttts:
Fixes https://github.com/openshift/origin/issues/20144

### DIFF
--- a/cmd/publishing-bot/publisher.go
+++ b/cmd/publishing-bot/publisher.go
@@ -114,11 +114,16 @@ func (p *PublisherMunger) ensureCloned(dst string, dstURL string) error {
 		return nil
 	}
 
-	err := exec.Command("mkdir", "-p", dst).Run()
-	if err != nil {
+	cmd := exec.Command("mkdir", "-p", dst)
+	if err := p.plog.Run(cmd); err != nil {
 		return err
 	}
-	cmd := exec.Command("git", "clone", "--no-tags", dstURL, dst)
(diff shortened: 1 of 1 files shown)


cmd/publishing-bot/publisher.go
```diff
@@ -114,11 +114,16 @@ func (p *PublisherMunger) ensureCloned(dst string, dstURL string) error {
...
-	if err != nil {
+	cmd := exec.Command("mkdir", "-p", dst)
+	if err := p.plog.Run(cmd); err != nil {
+		return err
+	}
+	cmd = exec.Command("git", "clone", dstURL, dst)
+	if err := p.plog.Run(cmd); err != nil {
 		return err
 	}
-	cmd := exec.Command("git", "clone", "--no-tags", dstURL, dst)
```
liggitt: was the intent to avoid pulling down content we don't need? can we upgrade the version of git in the image instead?

sttts: > can we upgrade the version of git in the image instead?

I would rather avoid pulling in any non-standard package. Our use of git is probably fragile. We are on a Debian LTS release for a reason.

> was the intent to avoid pulling down content we don't need?

We don't want `v1.11.0` like tags. Instead we fetch them "manually" in a controlled way and rewrite them to `kubernetes-1.11.0`.

liggitt:
ok

/lgtm

PR #74, opened 2018-07-30
Update to go 1.10.2

nikhita:
The publishing bot is currently broken because `strings.Builder is undefined`: https://github.com/kubernetes/kubernetes/issues/56876#issuecomment-408741947.

`strings.Builder` was introduced in Go 1.10. We use 1.10.2 in k/k (https://github.com/kubernetes/kubernetes/pull/63412, https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#external-dependencies).

This PR updates the go version to 1.10.2 for publishing bot so that it starts working again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,8 @@ RUN apt-get update \
  && apt-get install -y -qq git=1:2.1.4-2.1+deb8u5 \
  && apt-get install -y -qq mercurial \
  && apt-get install -y -qq ca-certificates wget jq vim tmux bsdmainutils tig \
- && wget https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz \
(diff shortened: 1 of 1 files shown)


nikhita:
/cc @sttts @mfojtik 

is this enough?

sttts:
This also shows a problem: we only have one Go version in the bot. So our smoke tests are bound to that Go version for all Kube versions.

zhangxiaoyu-zidif:
should we use the latest go version every time when compile K/K

nikhita:
> is that necessary that we use the latest go version every time when compile K/K

k/k's master uses 1.10.2 now so we should use that here. https://github.com/kubernetes/kubernetes/issues/67872 is the issue to move to 1.11.

> This also shows a problem: we only have one Go version in the bot. So our smoke tests are bound to that Go version for all Kube versions.

But we should ideally have support for multiple go versions.

PR #77, opened 2018-07-31
Fix 1.10 go cache

sttts:


artifacts/manifests/podspec.yaml
```diff
@@ -22,6 +22,8 @@ initContainers:
     name: publisher-gopath
   - mountPath: /etc/publisher-rules
     name: publisher-rules
+  - mountPath: /.cache
```
nikhita: why not just `/cache`?

sttts: Ask the golang devs ;-)

PR #80, opened 2018-08-06
Add branch rule go version value

sttts:
Kube 1.9 and 1.10 need Go 1.9, especially because go vet was extended in Go 1.10. So we better make this configurable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,13 +17,12 @@ MAINTAINER Chao Xu <xuchao@google.com>
 RUN apt-get update \
  && apt-get install -y -qq git=1:2.1.4-2.1+deb8u6 \
  && apt-get install -y -qq mercurial \
- && apt-get install -y -qq ca-certificates wget jq vim tmux bsdmainutils tig \
- && wget https://storage.googleapis.com/golang/go1.10.2.linux-amd64.tar.gz \
- && tar -C /usr/local -xzf go1.10.2.linux-amd64.tar.gz \
+ && apt-get install -y -qq ca-certificates curl wget jq vim tmux bsdmainutils tig \
  && rm -rf /var/lib/apt/lists/*
 
 ENV GOPATH="/go-workspace"
-ENV PATH="${GOPATH}/bin:/usr/local/go/bin:${PATH}"
+ENV GOROOT="/go-workspace/go"
+ENV PATH="${GOPATH}/bin:/go-workspace/go/bin:${PATH}"
 ENV GIT_COMMITTER_NAME="Kubernetes Publisher"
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -18,11 +18,18 @@ RUN apt-get update \
  && apt-get install -y -qq git=1:2.1.4-2.1+deb8u6 \
  && apt-get install -y -qq mercurial \
  && apt-get install -y -qq ca-certificates wget jq vim tmux bsdmainutils tig \
- && wget https://storage.googleapis.com/golang/go1.10.2.linux-amd64.tar.gz \
- && tar -C /usr/local -xzf go1.10.2.linux-amd64.tar.gz \
+ && for v in 1.9.7 1.10.2; do \
+      wget https://storage.googleapis.com/golang/go$v.linux-amd64.tar.gz && \
+      tar -C /usr/local -xzf go$v.linux-amd64.tar.gz && \
+      mv /usr/local/go /usr/local/go-$v; \
```
yue9944882:
```bash
ln -s /user/local/go-$v /user/local/go-${v%.*}
```
Might be cleaner here?

yue9944882: > mv /usr/local/go /usr/local/go-$v;

nit: do we need the tailing ';' ?

sttts: yes, it sperates the `done`.

cmd/init-repo/main.go
```diff
@@ -97,6 +98,31 @@ func main() {
 		glog.Fatalf("Failed to load rules: %v", err)
 	}
 
+	goVersions := []string{DefaultGoVersion}
```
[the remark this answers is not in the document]
sttts: Do we have sets here dependency-wise? The number of versions will be a hand full max. So the complexity of this quadratic loop is not an issue.

cmd/init-repo/main.go
```diff
@@ -97,6 +98,31 @@ func main() {
...
+				}
+			}
+		}
+	}
+	for _, v := range goVersions {
+		installGoVersion(v, filepath.Join(SystemGoPath, "go-"+v))
+	}
+	goLink, target := filepath.Join(SystemGoPath, "go"), filepath.Join(SystemGoPath, "go-"+DefaultGoVersion)
+	os.Remove(goLink)
+	if err := os.Symlink(target, goLink); err != nil {
```
yue9944882: @sttts  just to make sure: we are not supporting parallel build, am i right? 😃

sttts: No, we don't. But even then, this is just the default Go if no version is specified.

The whole go-workspace directory is a scratch pad for the publishing bot. I.e. it is used exclusively by one bot.

PR #82, opened 2018-08-07
Report better logs for github issue

yue9944882:
@sttts 

Plus some minor refactor 😃

Showing log in a better way for the bot.


cmd/publishing-bot/github.go
```diff
@@ -53,7 +53,20 @@ func ReportOnIssue(e error, logs, token, org, repo string, issue int) error {
 	}
 
 	// create new newComment
-	body := fmt.Sprintf("The last publishing run failed: %v\n\n```%s```\n", e, tail(logs, 65000))
+	// body := fmt.Sprintf("The last publishing run failed: %v\n\n```%s```\n", e, tail(logs, 65000))
```
sttts: no need for the old code, just drop the line

cmd/publishing-bot/log_builder.go
```diff
@@ -0,0 +1,93 @@
...
+			logBuilder.logs = append(logBuilder.logs, rawLogs[i])
+			break
+		}
+		logBuilder.logs = append(logBuilder.logs, rawLogs[i])
+		size += len(rawLogs)
+	}
+	return logBuilder
+}
+
+func (builder *logBuilder) addHeading(line string) *logBuilder {
```
sttts: would make the methods which are supposed to be used upper-case

cmd/publishing-bot/github.go
```diff
@@ -53,7 +53,20 @@ func ReportOnIssue(e error, logs, token, org, repo string, issue int) error {
...
+
+	body := newLogBuilderWithMaxBytes(65000, logs).
+		addHeading(fmt.Sprintf("The last publishing run failed: %v", e)).
+		addHeading("```").
+		trim("\n").
+		split("\n").
+		filter(func(line string) bool {
```
sttts: does this mean that all non-`+` lines are skipped?

yue9944882: @sttts yep. isn't that the behavior we want?

yue9944882: the skipped lines could be found in the file logs

sttts: Instead we want to show up to 50 `+` lines from the bottom, including all non-`+` lines. Above that we only show 50 non-`+` lines.

yue9944882: got it

cmd/publishing-bot/github.go
```diff
@@ -109,49 +110,26 @@ func CloseIssue(token, org, repo string, issue int) error {
...
-				return "..." + lines[0][max(0, len(lines[0])-maxBytes+3):]
+func transfromLogToGithubFormat(original string, maxLines int, headings ...string) string {
+	logCount := 0
+	transformed := NewLogBuilderWithMaxBytes(65000, original).
+		AddHeading(headings...).
+		AddHeading("```").
+		Trim("\n").
+		Split("\n").
+		Reverse().
+		Filter(func(line string) bool {
```
sttts: is this working backwards, i.e. we get the last lines?

yue9944882: yeah. it's a little tricky here. i did a reverse before filtering and reverse again to restore the order.

yue9944882: plz see the test i wrote for it

PR #83, opened 2018-08-09
cmd/*: add github-host and package flag

zhouhaibing089:
In case people are running github on-premise, they should be able
to tweak the github address.

the second case is: people might not be using the traditional name
convention like github.com/xxx/yyy as the package base, we should
allow them to specify the package name when cloning the code into
GOPATH.

### DIFF
--- a/cmd/init-repo/main.go
+++ b/cmd/init-repo/main.go
@@ -38,6 +38,9 @@ Command line flags override config values.
 
 func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
+	githubHost := flag.String("github-host", "", "the address of github (defaults to github.com)")
+	basePackage := flag.String("base-package", "", "the name of the package base (defaults to k8s.io when source repo is kubernetes, "+
+		"otherwise github-host/target-org)")
 	repoName := flag.String("source-repo", "", "the name of the source repository (eg. kubernetes)")
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	rulesFile := flag.String("rules-file", "", "the file with repository rules")
@@ -68,10 +71,26 @@ func main() {
 	if *repoOrg != "" {
 		cfg.SourceOrg = *repoOrg
 	}
-
-	if cfg.SourceRepo != "kubernetes" {
-		BaseRepoPath = filepath.Join(SystemGoPath, "src", "github.com", cfg.TargetOrg)
+	if *githubHost != "" {
+		cfg.GithubHost = *githubHost
 	}
+	if *basePackage != "" {
+		cfg.BasePackage = *basePackage
+	}
+
+	if cfg.GithubHost == "" {
+		cfg.GithubHost = "github.com"
+	}
+	// defaulting when base package is not specified
+	if cfg.BasePackage == "" {
+		if cfg.SourceRepo == "kubernetes" {
+			cfg.BasePackage = "k8s.io"
+		} else {
+			cfg.BasePackage = filepath.Join(cfg.GithubHost, cfg.TargetOrg)
+		}
+	}
+
+	BaseRepoPath = filepath.Join(SystemGoPath, "src", cfg.BasePackage)
 
 	if *rulesFile != "" {
 		cfg.RulesFile = *rulesFile
@@ -164,7 +183,7 @@ func installGoVersion(v string, pth string) {
 }
 
 func cloneForkRepo(cfg config.Config, repoName string) {
-	forkRepoLocation := "https://github.com/" + cfg.TargetOrg + "/" + repoName
+	forkRepoLocation := fmt.Sprintf("https://%s/%s/%s", cfg.GithubHost, cfg.TargetOrg, repoName)
 	repoDir := filepath.Join(BaseRepoPath, repoName)
 
 	if _, err := os.Stat(repoDir); err == nil {
@@ -246,7 +265,7 @@ func cloneSourceRepo(cfg config.Config, runGodepRestore bool) {
 		return
 	}
 
-	repoLocation := "https://" + filepath.Join("github.com", cfg.SourceOrg, cfg.SourceRepo)
+	repoLocation := fmt.Sprintf("https://%s/%s/%s", cfg.GithubHost, cfg.SourceOrg, cfg.SourceRepo)
 	glog.Infof("Cloning source repository %s ...", repoLocation)
 	cloneCmd := exec.Command("git", "clone", repoLocation)
 	run(cloneCmd)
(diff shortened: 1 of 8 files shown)


cmd/publishing-bot/config/config.go
```diff
@@ -18,6 +18,14 @@ package config
 
 // Config is how we are configured to talk to github.
 type Config struct {
+	// GithubHost is the address for github.
+	// Default to github.com
+	GithubHost string `yaml:"github-host"`
+
+	// Package is the package name for this repo.
+	// Default to k8s.io
+	Package string `yaml:"package"`
```
sttts: Wouldn't something like BasePackage be better? We have many packages around. But this is the base of all of them.

cmd/init-repo/main.go
```diff
@@ -68,11 +70,21 @@ func main() {
 	if *repoOrg != "" {
 		cfg.SourceOrg = *repoOrg
 	}
+	if *githubHost != "" {
+		cfg.GithubHost = *githubHost
+	}
+	if *packageName != "" {
+		cfg.Package = *packageName
```
sttts: this logic is strange. Shouldn't we preserve this if set, i.e. skip the following if clause?

cmd/init-repo/main.go
```diff
@@ -38,6 +38,8 @@ Command line flags override config values.
 
 func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
+	githubHost := flag.String("github-host", "github.com", "the address of github (eg. github.com)")
+	packageName := flag.String("package", "", "the name of the package base (eg. k8s.io)")
```
sttts: base-package would be better.

cmd/publishing-bot/publisher.go
```diff
@@ -137,7 +137,7 @@ func (p *PublisherMunger) construct() error {
 
 		// clone the destination repo
 		dstDir := filepath.Join(p.baseRepoPath, repoRule.DestinationRepository, "")
-		dstURL := fmt.Sprintf("https://github.com/%s/%s.git", p.config.TargetOrg, repoRule.DestinationRepository)
+		dstURL := fmt.Sprintf("https://%s/%s/%s.git", p.config.GithubHost, p.config.TargetOrg, repoRule.DestinationRepository)
```
sttts: while at it: can we unify the way we build the url? We have the `"x" + "/" + y + "/"` style, the `filepath.Join` style and now this one. This one is probably the most correct and idiomatic.

cmd/init-repo/main.go
zhouhaibing089: Good suggestion!

cmd/publishing-bot/config/config.go
zhouhaibing089: Updated to be `BasePackage`.

cmd/init-repo/main.go
zhouhaibing089: Oh! That's a good catch. The following `if` should only be run when `cfg.Package == ""`.

cmd/publishing-bot/publisher.go
zhouhaibing089: @sttts: I agree that we should rewrite `x + "/" + y + "/" +z` with `fmt.Sprintf` if there are more than three terms. 

`filepath.Join` is different though, it needs to handle `os.PathSeparator`.

cmd/init-repo/main.go
```diff
@@ -38,6 +38,8 @@ Command line flags override config values.
 
 func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
+	githubHost := flag.String("github-host", "", "the address of github (eg. github.com)")
```
sttts: s/eg/defaults to/

cmd/init-repo/main.go
```diff
@@ -68,10 +70,26 @@ func main() {
...
+	}
+	if *basePackage != "" {
+		cfg.BasePackage = *basePackage
+	}
 
-	if cfg.SourceRepo != "kubernetes" {
-		BaseRepoPath = filepath.Join(SystemGoPath, "src", "github.com", cfg.TargetOrg)
+	if cfg.GithubHost == "" {
+		cfg.GithubHost = "github.com"
 	}
```
sttts: just `} else` will do

cmd/init-repo/main.go
```diff
@@ -38,6 +38,8 @@ Command line flags override config values.
 
 func main() {
 	configFilePath := flag.String("config", "", "the config file in yaml format")
+	githubHost := flag.String("github-host", "", "the address of github (eg. github.com)")
+	basePackage := flag.String("base-package", "", "the name of the package base (eg. k8s.io)")
```
sttts: "defaults to k8s.io if source-repo is kubernetes and <github-host>/<target-org> otherwise"

cmd/init-repo/main.go
```diff
@@ -68,10 +70,26 @@ func main() {
...
+	if *githubHost != "" {
+		cfg.GithubHost = *githubHost
+	}
+	if *basePackage != "" {
+		cfg.BasePackage = *basePackage
+	}
 
-	if cfg.SourceRepo != "kubernetes" {
-		BaseRepoPath = filepath.Join(SystemGoPath, "src", "github.com", cfg.TargetOrg)
+	if cfg.GithubHost == "" {
```
sttts: put this into an else above

zhouhaibing089: @sttts: if we put this in the `else` branch above, there is no way to specify the value in config file.

sttts: sorry, misread.

cmd/publishing-bot/config/config.go
```diff
@@ -18,6 +18,14 @@ package config
 
 // Config is how we are configured to talk to github.
 type Config struct {
+	// GithubHost is the address for github.
+	// Default to github.com
+	GithubHost string `yaml:"github-host"`
+
+	// BasePackage is the base package name for this repo.
+	// Default to k8s.io
```
sttts: only for SourceOrg kubernetes. Otherwise, it is github.com/SourceOrg, isn't it?

cmd/init-repo/main.go
zhouhaibing089: fixed.

cmd/init-repo/main.go
zhouhaibing089: fixed.

cmd/publishing-bot/config/config.go
zhouhaibing089: correct!

sttts:
/approve
lgtm.

@mfojtik can you take a final look?

zhouhaibing089:
Kindly ping @mfojtik

zhouhaibing089:
@sttts: I add one more commit to use `base_package` in publishing scripts.

PR #87, opened 2018-08-17
init-repo: set Dir for go installation

zhouhaibing089:
the cmd to install go version should setup a dir explictly, or it
would fail like `chdir /go-workspace/x: no such file or directory`.

1. `run` setup `Dir` to be `BaseRepoPath` if it is empty at https://github.com/kubernetes/publishing-bot/blob/master/cmd/init-repo/main.go#L255
1. the invocation of [installGoVersion](https://github.com/kubernetes/publishing-bot/blob/master/cmd/init-repo/main.go#L137) sits before the [mkdir of BaseRepoPath](https://github.com/kubernetes/publishing-bot/blob/master/cmd/init-repo/main.go#L145)

### DIFF
--- a/cmd/init-repo/main.go
+++ b/cmd/init-repo/main.go
@@ -176,7 +176,9 @@ func installGoVersion(v string, pth string) {
 		glog.Fatal(err)
 	}
 	defer os.RemoveAll(tmpPath)
-	run(exec.Command("/bin/bash", "-c", fmt.Sprintf("curl -SLf https://storage.googleapis.com/golang/go%s.linux-amd64.tar.gz | tar -xz --strip 1 -C %s", v, tmpPath)))
+	cmd := exec.Command("/bin/bash", "-c", fmt.Sprintf("curl -SLf https://storage.googleapis.com/golang/go%s.linux-amd64.tar.gz | tar -xz --strip 1 -C %s", v, tmpPath))
+	cmd.Dir = tmpPath
+	run(cmd)
 	if err := os.Rename(tmpPath, pth); err != nil {
 		glog.Fatal(err)
 	}


cmd/init-repo/main.go
```diff
@@ -176,7 +176,9 @@ func installGoVersion(v string, pth string) {
 		glog.Fatal(err)
 	}
 	defer os.RemoveAll(tmpPath)
-	run(exec.Command("/bin/bash", "-c", fmt.Sprintf("curl -SLf https://storage.googleapis.com/golang/go%s.linux-amd64.tar.gz | tar -xz --strip 1 -C %s", v, tmpPath)))
+	cmd := exec.Command("/bin/bash", "-c", fmt.Sprintf("curl -SLf https://storage.googleapis.com/golang/go%s.linux-amd64.tar.gz | tar -xz --strip 1 -C %s", v, tmpPath))
+	cmd.Dir = tmpPath
+	run(cmd)
```
yue9944882: ain't that supposed to be created on `ioutil.TempDir(SystemGoPath, "go-tmp-")`?

zhouhaibing089: The point here is `cmd.Dir` must be present, so `tmpPath` is a choice because it is surely present since L174 creates it.

sttts: Isn't the `-C <tmpPath>` enough for tar to use it as destiniation?

zhouhaibing089: Sorry that I am not clear here, see this simplified example:

```console
$ cat main.go
package main

import (
        "os/exec"
        "fmt"
)

func main() {
        cmd := exec.Command("echo", "Hello World")
        cmd.Dir = "/some/path/does/not/exist"

        err := cmd.Run()
        if err != nil {
                fmt.Printf("%s\n", err)
        }
}
$ go run main.go
chdir /some/path/does/not/exist: no such file or directory
```

In the function here, `BaseRepoPath` specified [here](https://github.com/kubernetes/publishing-bot/blob/master/cmd/init-repo/main.go#L255) has not be created yet.

yue9944882: okay.. i got your point.. i presume you don't have `$GOPATH/src/k8s.io` present in your local environment, amiright?

[https://github.com/kubernetes/publishing-bot/blob/bb4df0689eb0de9e99cdb0280ef3ab6b59d2826b/cmd/init-repo/main.go#L255](https://github.com/kubernetes/publishing-bot/blob/bb4df0689eb0de9e99cdb0280ef3ab6b59d2826b/cmd/init-repo/main.go#L255)

zhouhaibing089: And `BaseRepoPath` will be created(`mkdir`) at https://github.com/kubernetes/publishing-bot/blob/master/cmd/init-repo/main.go#L145.

yue9944882: move these lines above `installGoVersion` would be much simplier, wdyt?

[https://github.com/kubernetes/publishing-bot/blob/bb4df0689eb0de9e99cdb0280ef3ab6b59d2826b/cmd/init-repo/main.go#L145-L147](https://github.com/kubernetes/publishing-bot/blob/bb4df0689eb0de9e99cdb0280ef3ab6b59d2826b/cmd/init-repo/main.go#L145-L147)

zhouhaibing089: Technically, I'm ok to move it before `installGoVersion`.

But honestly, I think it only makes sense to create directory `BaseRepoPath` when it is required,   and it seems `installGoVersion` does not require that path being present.

Let me know if you you would insist on doing that move.

sttts: I think both ways are fine.

PR #93, opened 2018-09-01
Add csi-api repo to publish bot

saad-ali:
We have a new https://github.com/kubernetes/csi-api repo that should be populated from https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/csi-api

I would like to turn on auto publishing.

/assign @sttts

### DIFF
--- a/configs/kubernetes-rules-configmap.yaml
+++ b/configs/kubernetes-rules-configmap.yaml
@@ -713,3 +713,17 @@ data:
           branch: release-1.11
         - repository: client-go
           branch: release-8.0
+    - destination: csi-api
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/csi-api
+        name: master
+        dependencies:
+        - repository: apimachinery
+          branch: master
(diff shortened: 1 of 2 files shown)


configs/kubernetes-rules-configmap.yaml
```diff
@@ -713,3 +713,15 @@ data:
...
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/csi-api
+        name: master
+        dependencies:
+        - repository: apimachinery
+          branch: master
+        - repository: api
```
sttts: no client-go?

sttts: godeps.json has client-go as well.

saad-ali: Ah will add

saad-ali:
Added client-go dependency. PTAL

sttts:
Smoke test is running to https://github.com/sttts/csi-api....

Issue #95, opened 2018-09-05
New branch forked before a staging dir was populated the first time break publishing


sttts:
That branch (1.12 in this case) was merged again and again with master, until the staging dir was created.

I assume that the filtered branch of the staging dir (here csi-api) does not have a proper corresponding commit for the fork point of the branch. We have to handle this case in addition.

PR #96, opened 2018-09-06
Handle branching points for new branches which do not contain subdir

sttts:
Fixes https://github.com/kubernetes/publishing-bot/issues/95.

We solve this by looking for the first (time-wise) fast-forward merge which introduced the subdirectory. The second-parent (on master) is the branching point for the new branch.

We do not support (yet) a publishing subdirectory that was actually created on a releaes branch.


sttts:
This is blocking Kubernetes 1.12 publishing.

sttts:
/hold

Smoketesting right now.

PR #97, opened 2018-09-07
Add new repositories for component config and cli

dims:
New repos requested in:
https://github.com/kubernetes/org/issues/58
https://github.com/kubernetes/org/issues/56

Change-Id: I2e71a998db2e53b8d8827a59a7b35198f52e3e13

### DIFF
--- a/configs/kubernetes-rules-configmap.yaml
+++ b/configs/kubernetes-rules-configmap.yaml
@@ -846,3 +846,79 @@ data:
           branch: release-1.12
         - repository: client-go
           branch: release-9.0
+    - destination: cli-runtime
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/cli-runtime
+        name: master
+        dependencies:
+        - repository: api
+          branch: master
+        - repository: apimachinery
+          branch: master
+        - repository: client-go
+          branch: master
+    - destination: sample-cli-plugin
+      library: false
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/sample-cli-plugin
+        name: master
+        dependencies:
+        - repository: api
+          branch: master
+        - repository: apimachinery
+          branch: master
+        - repository: cli-runtime
+          branch: master
+        - repository: client-go
+          branch: master
+    - destination: kube-proxy
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/kube-proxy
+        name: master
+        dependencies:
+        - repository: apimachinery
+          branch: master
+    - destination: kubelet
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/kubelet
+        name: master
+        dependencies:
+        - repository: apimachinery
+          branch: master
+        - repository: api
+          branch: master
+    - destination: kube-scheduler
+      library: true
+      branches:
+      - source:
+          branch: master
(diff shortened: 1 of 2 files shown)


configs/kubernetes-rules-configmap.yaml
```diff
@@ -846,3 +846,79 @@ data:
           branch: release-1.12
         - repository: client-go
           branch: release-9.0
+    - destination: sample-cli-plugin
+      library: true
```
sttts: is this a library? I.e. do we want complete Godeps? For libraries we remove some of the k8s.io dependencies to be go-gettable. Ugly but necessary.

configs/kubernetes-rules-configmap.yaml
```diff
@@ -846,3 +846,79 @@ data:
...
+          branch: master
+    - destination: kube-proxy
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/kube-proxy
+        name: master
+        dependencies:
+        - repository: apimachinery
```
sttts: api, client-go are missing

configs/kubernetes-rules-configmap.yaml
```diff
@@ -846,3 +846,79 @@ data:
...
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/kubelet
+        name: master
+        dependencies:
+        - repository: apimachinery
+          branch: master
+        - repository: api
```
sttts: client-go is missing

configs/kubernetes-rules-configmap.yaml
sttts: nvmd. With the config apis only right, porbably it is complete

configs/kubernetes-rules-configmap.yaml
sttts: nvmd. With the config apis only right, porbably it is complete

configs/kubernetes-rules-configmap.yaml
sttts: After looking into the staging repo, this is most certainly not a library, but a command only.

configs/kubernetes-rules-configmap.yaml
dims: y i checked with GoDeps

configs/kubernetes-rules-configmap.yaml
dims: yep, listed only those in GoDeps

configs/kubernetes-rules-configmap.yaml
dims: so i change this to `library: false`?

dims: Fixed. set to false

configs/kubernetes-rules-configmap.yaml
```diff
@@ -846,3 +846,79 @@ data:
...
+        dependencies:
+        - repository: api
+          branch: master
+        - repository: apimachinery
+          branch: master
+        - repository: cli-runtime
+          branch: master
+        - repository: client-go
+          branch: master
+    - destination: cli-runtime
```
sttts: order is wrong: it must be in topological order, i.e. depending repos after dependencies.

dims: Fixed

sttts:
Repos have been published. 🎉 Closing as the follow-up PR has been merged.

PR #101, opened 2018-09-07
Speed up if nothing changed on a branch

sttts:
- skip cherry-picking if published hash matches last upstream hash
- skip commit mapping computation if no new tag is found

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 13 ]; then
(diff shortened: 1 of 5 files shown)


artifacts/scripts/construct.sh
```diff
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 13 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_packages kubernetes_remote subdirectory source_repo_org source_repo_name base_package is_library recursive_delete_pattern skip_tags"
+if [ ! $# -eq 14 ]; then
```
mfojtik: i wonder if this ever stop :)

PR #103, opened 2018-09-08
Clean up updateSourceRepo and checkout branch before rule loading

sttts:
We did not checkout master before reading the rules file, potentially from the source repo.

Also the source repo branch logic gave dirty output.


sttts:
Smoke test has been green.

/assign @dims

PR #116, opened 2018-10-04
Fix merge point calculation for repeated merges with same branch commit

sttts:
A commit might be ancestor of multiple merges into the mainline. Only the first of those merges is considered the merge point of that commit. To find it we have to traverse mainline merges from oldest to newest, storing the first appearance of a commit as ancestor of a merge.

All credit goes to @nikhita for uncovering this.

### DIFF
--- a/pkg/git/mainline.go
+++ b/pkg/git/mainline.go
@@ -72,7 +72,7 @@ func MergePoints(r *gogit.Repository, mainLine []*object.Commit) (map[plumbing.H
 		mainLinePos[c.Hash] = i
 	}
 
-	bestMergePoints := map[plumbing.Hash]int{}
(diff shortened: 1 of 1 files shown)


pkg/git/mainline.go
```diff
@@ -128,7 +129,7 @@ func MergePoints(r *gogit.Repository, mainLine []*object.Commit) (map[plumbing.H
 	// map commit hash to best merge point on mainline
```
nikhita: nit: best/earliest

nikhita:
/lgtm
/hold

Please `/hold cancel` if you think we should get the commit message fix in another PR.

nikhita:
/hold cancel

from the discussion on slack, let's get the commit message fix in later

PR #117, opened 2018-10-04
Bump go-git to private branch to fix commit messages

sttts:
Compare https://github.com/src-d/go-git/pull/976.


Gopkg.lock
```diff
@@ -1,6 +1,21 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  digest = "1:b498b36dbb2b306d1c5205ee5236c9e60352be8f9eea9bf08186723a9f75b4f3"
+  name = "github.com/emirpasic/gods"
```
nikhita: why is this needed though?

sttts: Some dependency it seems.

PR #123, opened 2018-10-26
Handle degenerated non-mainline merge

sttts:
Handling merges on feature branches is tricky, especially when the feature branch
is degenerated to single commits on the filtered mainline. Here were add such a
merge commit that is very seldom as non-overlapping feature branches with
intermediate master merges do not happen often.

The respective merge is from https://github.com/kubernetes/kubernetes/pull/69946.

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -518,6 +518,7 @@ sync_repo() {
 function pick-merge-as-single-commit() {
     grep -F -q -x "$1" <<EOF
 25ebf875b4235cb8f43be2aec699d62e78339cec
+8014d73345233c773891f26008e55dc3b5232c7c
 EOF
 }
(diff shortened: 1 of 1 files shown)


artifacts/scripts/util.sh
```diff
@@ -518,6 +518,7 @@ sync_repo() {
 function pick-merge-as-single-commit() {
     grep -F -q -x "$1" <<EOF
 25ebf875b4235cb8f43be2aec699d62e78339cec
+8014d73345233c773891f26008e55dc3b5232c7c
```
nikhita: oh wow, didn't know we black listed it this way.

sttts:
/hold

smoke testing

nikhita:
Oops, prow is too fast with merges.

PR #125, opened 2018-11-01
README: remove tagging as a Known Issue

nikhita:
https://github.com/kubernetes/kubernetes/issues/49011 has also been closed.

/kind cleanup
/assign @sttts


nikhita:
> We don't automatically add new branches.

Ah, makes sense.
/close

PR #130, opened 2018-11-02
Recheckout when text conversion mode is changed

sttts:


mfojtik:
/lgtm

Provides a migration which is always good :)

sttts:
Smoke test is green of the initial source repo fetch.

mfojtik:
/hold cancel
/lgtm

Tested this and it does the job.

PR #132, opened 2018-11-08
Add 1.13 branches and disable publishing of 1.9

nikhita:
Ref: https://groups.google.com/d/topic/kubernetes-dev/8D_ea0xw3BQ/discussion

1.13 branch has been created in k/k and with that we stop supporting 1.9 and older.

/assign @sttts

### DIFF
--- a/configs/kubernetes-rules-configmap.yaml
+++ b/configs/kubernetes-rules-configmap.yaml
@@ -24,11 +24,11 @@ data:
       #     branch: release-1.8
       #     dir: staging/src/k8s.io/code-generator
       #   name: release-1.8
-      - source:
-          branch: release-1.9
-          dir: staging/src/k8s.io/code-generator
-        name: release-1.9
-        go: 1.9.7
+      # - source:
+      #    branch: release-1.9
+      #    dir: staging/src/k8s.io/code-generator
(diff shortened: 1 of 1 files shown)


configs/kubernetes-rules-configmap.yaml
```diff
@@ -211,6 +229,16 @@ data:
           branch: release-1.12
         - repository: api
           branch: release-1.12
+      - source:
+          branch: release-1.13
+          dir: staging/src/k8s.io/client-go
+        name: release-10.0
```
nikhita: Can't believe we are at v10.0.0 of client-go already :smile:

configs/kubernetes-rules-configmap.yaml
```diff
@@ -1037,3 +1243,14 @@ data:
           branch: master
         - repository: client-go
           branch: master
+      - source:
+          branch: release-1.13
+          dir: staging/src/k8s.io/cloud-provider
+        name: release-1.13
+        dependencies:
```
nikhita: Missed the go version here. Will add.

nikhita: Added.

PR #134, opened 2018-11-14
Fix bug: shell syntax error

SataQiu:
Fix bug: shell syntax error, ${1-k8s.io}  -> ${1:-k8s.io}

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -65,7 +65,7 @@ SOURCE_REPO_NAME="${9}"
 shift 9
 
 # base package name (eg. k8s.io)
-BASE_PACKAGE="${1-k8s.io}"
+BASE_PACKAGE="${1:-k8s.io}"
 # If ${REPO} is a library
 IS_LIBRARY="${2}"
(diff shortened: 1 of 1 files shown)


artifacts/scripts/construct.sh
```diff
@@ -65,7 +65,7 @@ SOURCE_REPO_NAME="${9}"
 shift 9
 
 # base package name (eg. k8s.io)
-BASE_PACKAGE="${1-k8s.io}"
+BASE_PACKAGE="${1:-k8s.io}"
```
nikhita: @SataQiu Thanks! 

Just to be clear, this is not a syntax error. :)

`${foo:bar}` expands to `bar` when foo is unset, but `""` when foo is null, while `${foo:-bar}` expands to `bar` if foo is unset or null.

nikhita:
/ok-to-test
/lgtm

/assign @sttts 
for approval

artifacts/scripts/construct.sh
SataQiu: @nikhita Thank you for your patient explanation. :)

PR #138, opened 2018-12-03
Dockerfile: add gcc dep due to cgo

sttts:


sttts:
/hold cancel

Smoke test is green.

PR #140, opened 2018-12-20
Switch to ReplicaSet

sttts:
ReplicationControllers have been replaced by RSs long ago.


nikhita:
Oh wow. The README also mentions "ReplicationController". We should change that as well. :)

sttts:
@nikhita Happy new year! Updated the README.

PR #142, opened 2019-01-04
Add component-base repo

nikhita:
- Repo creation request: https://github.com/kubernetes/org/issues/330
- Repo with initial empty commit: https://github.com/kubernetes/component-base
- PR against test-infra for branch protection: https://github.com/kubernetes/test-infra/pull/10635
- PR for staging against k/k: https://github.com/kubernetes/kubernetes/pull/72569
- Dependencies: As per PR [`Godeps.json`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/Godeps/Godeps.json),
  -  only apimachinery is a dependency.
  - `component-base` is a dependency to `kube-proxy`, `kube-scheduler` and `kube-controller-manager`.
  - it is safe to add future dependencies per https://github.com/kubernetes/publishing-bot/pull/142#discussion_r245485117 as well, so they have been added too.

/cc @luxas @sttts 
/assign @sttts 

/hold 

I'll consider it safe to cancel the hold when:
- [x] The published repo has been created and has an initial empty commit: https://github.com/kubernetes/component-base.
- [x] The PR against staging (https://github.com/kubernetes/kubernetes/pull/72569) has merged.
- [x] Dependencies are verified after the above PR is merged (just to double check). Note: future dependencies have also been added.
- [x] The PR against test-infra for branch protection is merged (https://github.com/kubernetes/test-infra/pull/10635)

### DIFF
--- a/configs/kubernetes-rules-configmap.yaml
+++ b/configs/kubernetes-rules-configmap.yaml
@@ -243,6 +243,20 @@ data:
         godep restore
         go build ./...
         go test $(go list ./... | grep -v /vendor/)
+    - destination: component-base
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/component-base
+        name: master
+        dependencies:
+        - repository: api
+          branch: master
+        - repository: apimachinery
+          branch: master
+        - repository: client-go
+          branch: master
     - destination: apiserver
       library: true
       branches:
@@ -256,6 +270,8 @@ data:
         - repository: api
           branch: master
         - repository: client-go
+          branch: master
+        - repository: component-base
           branch: master
       # - source:
       #     branch: release-1.6
@@ -360,6 +376,8 @@ data:
         - repository: client-go
           branch: master
         - repository: apiserver
+          branch: master
+        - repository: component-base
           branch: master
       # - source:
       #     branch: release-1.6
@@ -482,6 +500,8 @@ data:
         - repository: apiserver
           branch: master
         - repository: code-generator
+          branch: master
+        - repository: component-base
           branch: master
         required-packages:
         - k8s.io/code-generator
@@ -636,6 +656,8 @@ data:
         - repository: client-go
           branch: master
         - repository: code-generator
+          branch: master
+        - repository: component-base
           branch: master
         required-packages:
         - k8s.io/code-generator
@@ -1057,6 +1079,8 @@ data:
           branch: master
         - repository: client-go
           branch: master
+        - repository: component-base
+          branch: master
       - source:
           branch: release-1.12
           dir: staging/src/k8s.io/sample-cli-plugin
@@ -1094,6 +1118,8 @@ data:
         name: master
         dependencies:
         - repository: apimachinery
+          branch: master
+        - repository: component-base
           branch: master
       - source:
           branch: release-1.12
@@ -1155,6 +1181,8 @@ data:
           branch: master
         - repository: apiserver
           branch: master
+        - repository: component-base
+          branch: master
       - source:
           branch: release-1.12
           dir: staging/src/k8s.io/kube-scheduler
@@ -1186,6 +1214,8 @@ data:
         - repository: apimachinery
           branch: master
         - repository: apiserver
+          branch: master
+        - repository: component-base
           branch: master
       - source:
           branch: release-1.12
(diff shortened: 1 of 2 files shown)


configs/kubernetes-rules-configmap.yaml
```diff
@@ -1095,6 +1095,8 @@ data:
         dependencies:
         - repository: apimachinery
           branch: master
+        - repository: component-base
```
sttts: also apiserver and the different sample-* projects will depend on it.

configs/kubernetes-rules-configmap.yaml
```diff
@@ -1269,3 +1275,13 @@ data:
...
           branch: master
+    - destination: component-base
+      library: true
+      branches:
+      - source:
+          branch: master
+          dir: staging/src/k8s.io/component-base
+        name: master
+        dependencies:
+        - repository: apimachinery
```
sttts: it also depends on client-go and api.

configs/kubernetes-rules-configmap.yaml
nikhita: > will depend on it.

Is it safe to add component-base as a dependency to those repos even if it is not a dependency yet? For now, only kube-proxy, kube-scheduler and kube-controller-manager have it as a dependency (as per https://github.com/kubernetes/kubernetes/pull/72569)

configs/kubernetes-rules-configmap.yaml
nikhita: Same question as above ^, right now in the PR component-base only depends on apimachinery: https://github.com/kubernetes/kubernetes/blob/66d879c94637489c03d16651ba01a15e60b6116a/staging/src/k8s.io/component-base/Godeps/Godeps.json.

configs/kubernetes-rules-configmap.yaml
sttts: it only means that the order of publishing is accordingly. It does not change anything else.

luxas: Great, then lets do it

configs/kubernetes-rules-configmap.yaml
luxas: @sttts shall we wait until it actually depends on those two as well? or is this just (again) about publishing order?

hack/fetch-all-latest-and-push.sh
```diff
@@ -33,26 +33,27 @@ if [ "$#" -ge 2 ]; then
 fi
 GITHUB_HOST=${GITHUB_HOST:-github.com}
 repos=(
-    apimachinery
```
luxas: Thanks for sorting!

configs/kubernetes-rules-configmap.yaml
nikhita: :heavy_check_mark: Done.

configs/kubernetes-rules-configmap.yaml
nikhita: I have added here as well, but let me know if you'd like to wait.

> or is this just (again) about publishing order?

Looking at the code, it looks like the list of dependencies are just used to remove the values for those deps in `Godeps/Godeps.json`, and then regenerate them. If it doesn't find a dep, it'll be a no-op. 

https://github.com/kubernetes/publishing-bot/blob/54cd0ea15d70bd75141fc02405f2d8203fd3fbfa/artifacts/scripts/util.sh#L815-L819

But let's wait for @sttts to confirm this. :)

sttts: see https://github.com/kubernetes/publishing-bot/pull/142#discussion_r245485117

luxas:
@nikhita both https://github.com/kubernetes/kubernetes/pull/72569 and https://github.com/kubernetes/test-infra/pull/10635 have now merged, I think we should be good to go here :)

nikhita:
Updated the check list in the main PR body and cancelling the hold. :rocket: 

/hold cancel

@sttts can you approve?

nikhita:
I updated to make sure that `component-base` comes before `apiserver` (because apiserver depends on it, so it should come first in the list). And I also added `component-base` as a dependency to `kube-aggregator` because `kube-aggregator` _will_ depend on it later.

@luxas @sttts  Ptal.

luxas:
Why hold? Do you need some bugfix(es) to the bot merged first?

luxas:
ah I see

nikhita:
Closing in favour of https://github.com/kubernetes/publishing-bot/pull/144. These commits have been picked in https://github.com/kubernetes/publishing-bot/pull/144.

PR #153, opened 2019-01-25
Skip godeps generation for commits containing incorrect godep changes

nikhita:
This commit adds an exception list for commits containing bad godep changes
(eg client-go depending on apiserver) and skips godeps generation for these
commits.

Note that we will still reset godeps for each commit (including the bad commit).
And the correct godep generation will take place once it encounters a good commit.

This is a temporary fix and should probably be reverted because it is not always
true that there will be a commit having the correct godep changes after a bad one
_immediately_. If there isn't a good commit after a bad one immediately, the commits
between them will have bad Godeps.json.

---

This is the bad commit that needs to skip godeps generation - https://github.com/kubernetes/kubernetes/commit/e2a017327c1af628f4f0069cbd49865ad1e81975 and this is the good commit that reverts the bad change - https://github.com/kubernetes/kubernetes/commit/c8d89b34cc692c4b477fe40ac4511be73a581629.

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -336,7 +336,9 @@ sync_repo() {
             local dst_new_merge=$(GIT_COMMITTER_DATE="${date}" GIT_AUTHOR_DATE="${date}" git commit-tree -p ${dst_merge_point_commit} -p ${dst_parent2} -m "$(commit-message ${k_pending_merge_commit}; echo; echo "${commit_msg_tag}: ${k_pending_merge_commit}")" HEAD^{tree})
             # no amend-godeps needed here: because the merge-commit was dropped, both parents had the same tree, i.e. Godeps.json did not change.
             git reset -q --hard ${dst_new_merge}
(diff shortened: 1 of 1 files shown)


artifacts/scripts/util.sh
```diff
@@ -484,9 +484,11 @@ sync_repo() {
             # we would end up with "base B + change B" which misses the change A changes.
             amend-godeps-at ${f_mainline_commit}
 
-            fix-godeps "${deps}" "${required_packages}" "${base_package}" "${is_library}" ${dst_needs_godeps_update} true ${commit_msg_tag} "${recursive_delete_pattern}"
-            dst_needs_godeps_update=false
-            dst_merge_point_commit=$(git rev-parse HEAD)
+            if ! pick-incorrect-godep-changes ${k_mainline_commit}; then
```
nikhita: @sttts Should `amend-godeps-at` above be also skipped?

nikhita:
Squashed.

PR #157, opened 2019-02-01
Move golang download to construction time

sttts:
Necessary if the rules file comes from a repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ validate:
 .PHONY: validate
 
 init-deploy: validate
-	$(KUBECTL) delete -n "$(NAMESPACE)" --ignore-not-found=true rs publisher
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -50,7 +50,7 @@ validate:
 .PHONY: validate
 
 init-deploy: validate
-	$(KUBECTL) delete -n "$(NAMESPACE)" --ignore-not-found=true rs publisher
+	$(KUBECTL) delete -n "$(NAMESPACE)" --ignore-not-found=true replicaset publisher
```
dims: unrelated. but i am ok with merging it in this PR if you wish

sttts: rs should even work. It didn't, maybe for some completely unrelated reason.

PR #158, opened 2019-02-07
Fixes for running kubernetes-nightly config in CNCF cluster

dims:
- fix apiversion
- add missing file
- fix source-org
- run git config with user.email and user.name
- install default golang in init-repo itself

Change-Id: I102b0ae11aa141bc338fcfcea9b9913c600cc745

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -87,6 +87,10 @@ if [ ! -f .git/info/attributes ]; then
     # switch over to new file endings
     rm -r * && git checkout . || true
 fi
+
+git config user.email "$GIT_COMMITTER_EMAIL"
+git config user.name "$GIT_COMMITTER_NAME"
+
 echo "Running garbage collection."
 git gc --auto
 echo "Fetching from origin."
--- a/cmd/publishing-bot/publisher.go
+++ b/cmd/publishing-bot/publisher.go
@@ -31,6 +31,7 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing"
 
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
+	"k8s.io/publishing-bot/pkg/golang"
 )
 
 // PublisherMunger publishes content from one repository to another one.
@@ -195,7 +196,7 @@ func (p *PublisherMunger) ensureCloned(dst string, dstURL string) error {
 func (p *PublisherMunger) construct() error {
 	sourceRemote := filepath.Join(p.baseRepoPath, p.config.SourceRepo, ".git")
 
-	if err := installGoVersions(&p.reposRules); err != nil {
+	if err := golang.InstallGoVersions(&p.reposRules); err != nil {
 		return err
 	}
 
--- a/configs/kubernetes-nightly-configmap.yaml
+++ b/configs/kubernetes-nightly-configmap.yaml
@@ -1,9 +1,10 @@
+apiVersion: v1
 kind: ConfigMap
 metadata:
   name: publisher-config
 data:
   config: |
-    source-org: kubernetes-nightly
+    source-org: kubernetes
     source-repo: kubernetes
     target-org: kubernetes-nightly
     rules-file: https://raw.githubusercontent.com/kubernetes/kubernetes/master/staging/publishing/rules.yaml
--- a/cmd/init-repo/main.go
+++ b/cmd/init-repo/main.go
@@ -13,6 +13,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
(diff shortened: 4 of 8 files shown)


configs/kubernetes-nightly-configmap.yaml
```diff
@@ -1,3 +1,4 @@
+apiVersion: v1
```
nikhita: ouch :)

Dockerfile
```diff
@@ -12,19 +12,23 @@
...
  && apt-get install -y -qq ca-certificates curl wget jq vim tmux bsdmainutils tig gcc \
  && rm -rf /var/lib/apt/lists/*
 
-ENV GOPATH="/go-workspace"
-ENV GOROOT="/go-workspace/go"
-ENV PATH="${GOPATH}/bin:/go-workspace/go/bin:${PATH}"
 ENV GIT_COMMITTER_NAME="Kubernetes Publisher"
 ENV GIT_COMMITTER_EMAIL="k8s-publishing-bot@users.noreply.github.com"
+RUN git config --global user.email "Kubernetes Publisher"
+RUN git config --global user.name "k8s-publishing-bot@users.noreply.github.com"
```
sttts: Wondering why we used env var. It failed without?

configs/kubernetes-nightly-configmap.yaml
```diff
@@ -1,9 +1,10 @@
...
 kind: ConfigMap
 metadata:
   name: publisher-config
 data:
   config: |
-    source-org: kubernetes-nightly
+    source-org: kubernetes
```
sttts: the source-org would still be kubernetes I believe. Otherwise, import paths do not work.

Dockerfile
```diff
@@ -12,19 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:stretch
+FROM golang:1.11-stretch
```
sttts: Is this necessary? I would prefer full control over the go version. No go version anywhere in the path.

configs/kubernetes-nightly-configmap.yaml
dims: agree

Dockerfile
dims: i will probably drop this and move it into code

Dockerfile
dims: ack. will play with this image and then try going back to debian:stretch

dims: @sttts in init-repo/main.go, we try to install `github.com/tools/godep` and `github.com/golang/dep` for which we need a golang present in the container. right?

dims:
/assign @sttts @nikhita 

This is ready. got a full run in the CNCF cluster using the kubernetes-nightly config

dims:
need this to work well in the CNCF cluster: https://github.com/kubernetes/k8s.io/issues/183

artifacts/scripts/util.sh
```diff
@@ -776,6 +776,9 @@ function reset-godeps() {
         git rm -rf Godeps
     fi
 
+    git config user.email "$GIT_COMMITTER_EMAIL"
```
sttts: can't we set this once?

cmd/publishing-bot/publisher.go
```diff
@@ -287,6 +288,8 @@ func (p *PublisherMunger) construct() error {
 				lastPublishedUpstreamHash,
 			)
 			cmd.Env = append([]string(nil), branchEnv...) // make mutable
+			cmd.Stdout = os.Stdout
```
sttts: why this?

artifacts/scripts/util.sh
dims: i could not figure out where we could. any suggestions? i threw them in only when they failed. so took multiple iterations to get this set of changes

cmd/publishing-bot/publisher.go
dims: i will remove this.

artifacts/scripts/util.sh
sttts: Why not move it to https://github.com/kubernetes/publishing-bot/blob/master/artifacts/scripts/construct.sh#L83 ?

artifacts/scripts/construct.sh
```diff
@@ -86,6 +86,8 @@ if [ ! -f .git/info/attributes ]; then
     echo "* -text" >> .git/info/attributes
     # switch over to new file endings
     rm -r * && git checkout . || true
+    git config user.email "$GIT_COMMITTER_EMAIL"
```
sttts: outside of the if clause.

dims: 👍

artifacts/scripts/util.sh
sttts: this is called from construct.sh only. No need for these two lines.

PR #161, opened 2019-02-11
Make validate-rules command a bit more verbose

dims:
Change-Id: I3e9c8eb2542b50f81ba7cc1ea594f2a12384f9dd

### DIFF
--- a/cmd/publishing-bot/config/rules.go
+++ b/cmd/publishing-bot/config/rules.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/glog"
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/publishing-bot/config/rules.go
```diff
@@ -118,6 +120,7 @@ func readFromUrl(u *url.URL) ([]byte, error) {
 }
 
 func validateRepoOrder(rules *RepositoryRules) (errs []error) {
+	fmt.Printf("validating repository order\n")
```
sttts: we use glog elsewhere if I am not mistaken.

dims: ack i can switch it over

dims: Done

PR #164, opened 2019-03-04
Bump default go version to 1.12

nikhita:
Ref: https://github.com/kubernetes/kubernetes/pull/74632

/cc @dims @sttts

### DIFF
--- a/pkg/golang/install.go
+++ b/pkg/golang/install.go
@@ -29,7 +29,7 @@ import (
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
 )
 
-const defaultGoVersion = "1.11.5"
+const defaultGoVersion = "1.12"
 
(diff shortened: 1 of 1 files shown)


sttts:
Have we hardcoded the version everywhere else, so that this is has no consequences?

nikhita:
> Have we hardcoded the version everywhere else

No, we don't hard code the default go version anywhere else.

The only other place we hardcore the go version is in the rules. But the rules don't hard code the go version for the master branch - only the release branches. The `defaultGoVersion` is used for the master branch.

And the go version for all 1.14 branches has been updated to 1.12 in https://github.com/kubernetes/kubernetes/pull/74664. :+1: 

https://github.com/kubernetes/kubernetes/blob/master/staging/publishing/rules.yaml

PR #166, opened 2019-03-11
Retry quickly when references changed

dims:
Lets not wait 4 hours and retry quickly. If this works, we can add another parameter instead of hardcoding 5 mins.

Change-Id: I2ab92772f417124c4684bb5fb17bc5437923812d

### DIFF
--- a/cmd/publishing-bot/main.go
+++ b/cmd/publishing-bot/main.go
@@ -23,6 +23,7 @@ import (
 	"os"
 
 	"github.com/golang/glog"
+	"gopkg.in/src-d/go-git.v4/storage/memory"
 	"gopkg.in/yaml.v2"
 
 	"strings"
@@ -163,6 +164,7 @@ func main() {
 	}
 
 	for {
+		waitfor := *interval
 		last := time.Now()
 		publisher := New(&cfg, baseRepoPath)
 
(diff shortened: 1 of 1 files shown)


cmd/publishing-bot/main.go
```diff
@@ -183,6 +185,11 @@ func main() {
 					githubIssueErrorf("Failed to report logs on github issue: %v", err)
 					server.SetHealth(false, hash)
 				}
+				if strings.HasSuffix(err.Error(), memory.ErrRefHasChanged.Error()) {
+					// If the issue is just "reference has changed concurrently",
+					// then let us wait for 5 minutes and try again.
```
nikhita: Let's also add some logging here...also, can we add a comment here with a TODO to fix the actual flake?

nikhita: Can this end up in some weird recursion? :upside_down_face: 

We have seen this error consecutively several times...maybe we should let it retry only a certain number of times?

dims: done.

dims: i was thinking we should try this as-is and we can remove this if it does not work.

nikhita:
Why do we have travis CI run here? It was removed in https://github.com/kubernetes/publishing-bot/pull/128.....

nikhita:
We have prow now which runs `make` so I'll remove the webhook for travis.

https://github.com/kubernetes/test-infra/blob/a967141fc1ac3faee4e703dcf8ba4201f17c9cfa/config/jobs/kubernetes/publishing-bot/publishing-bot-presubmits.yaml#L2-L11

nikhita:
Closing and reopening to remove travis.

nikhita:
O_o

nikhita:
Travis didn't run any of the past few PRs...why this one. thinking:

PR #169, opened 2019-03-21
go-mod support

sttts:
This removes godeps support and adds generation of go.mod+sum. No vendor/ is generated anymore.

In parallel, there is a godeps branch forked off before this PR. We will keep running a second instance for all pre-1.15 versions.

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -367,7 +367,12 @@ sync_repo() {
             local squash_commits=1
             if gomod-changes ${f_mainline_commit}; then
                 reset-gomod ${f_mainline_commit}^
-                squash_commits=2 # squash the cherry-pick into the go.mod reset commit below
+                squash_commits=$[${squash_commits} + 1] # squash the cherry-pick into the go.mod reset commit below
+                dst_needs_gomod_update=true
+            fi
+            if godeps-changes ${f_mainline_commit}; then
+                reset-godeps ${f_mainline_commit}^
+                squash_commits=$[${squash_commits} + 1]
                 dst_needs_gomod_update=true
             fi
 
@@ -425,7 +430,12 @@ sync_repo() {
                 local squash_commits=1
                 if gomod-changes ${f_latest_branch_point_commit} ${f_latest_merge_commit}; then
                     reset-gomod ${f_latest_branch_point_commit}
-                    squash_commits=2 # squash the cherry-pick into the go.mod reset commit below
+                    squash_commits=$[${squash_commits} + 1] # squash the cherry-pick into the go.mod reset commit below
+                    dst_needs_gomod_update=true
+                fi
+                if godeps-changes ${f_latest_branch_point_commit} ${f_latest_merge_commit}; then
+                    reset-godeps ${f_latest_branch_point_commit}
+                    squash_commits=$[${squash_commits} + 1]
                     dst_needs_gomod_update=true
                 fi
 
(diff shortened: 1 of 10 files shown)


artifacts/scripts/util.sh
```diff
@@ -702,48 +702,20 @@ function fix-gomod() {
...
-
-        # check if there are new contents
-        if ! git-index-clean; then
-           echo "Committing vendor/ with required packages: ${required_packages}"
-           git commit -q -m "sync: update required packages"
-        fi
-    fi
-
-    # required packages above could have added files to be deleted according to delete pattern
-    apply-recursive-delete-pattern "${recursive_delete_pattern}"
```
liggitt: is this still needed to get rid of BUILD files?

artifacts/scripts/util.sh
```diff
@@ -787,169 +759,47 @@ function squash() {
...
-        elif git grep -w -q "k8s.io/${dep}"; then
-            # revert changes and fall back to full vendoring
-            echo "Found new dependency k8s.io/${dep}. Switching to full vendoring."
-            git checkout -q HEAD Godeps/Godeps.json
-            update_full_gomod "${deps}" "${base_package}" "${is_library}" "${commit_msg_tag}"
-            return $?
-        else
-            echo "Ignoring k8s.io/${dep} dependency because it seems not to be used."
-        fi
+        local dep_commit=$(cd ../${dep}; git show -q --pretty='format:v0.0.0-%cd-%H' --date='format:%Y%m%d%H%M%S')
```
liggitt: using %h and --abbrev=12 gets the format we want:

```
git show -q --pretty='format:v0.0.0-%cd-%h' --date='format:%Y%m%d%H%M%S' --abbrev=12
```

sttts: fixed

artifacts/scripts/util.sh
```diff
@@ -990,6 +904,26 @@ checkout-deps-to-kube-commit() {
         pushd ../${dep} >/dev/null
             echo "Checking out k8s.io/${dep} to ${dep_commit}"
             git checkout -q "${dep_commit}"
+
+            echo "Downloading go mod dependencies..."
+            go mod download
+
+            local pseudo_version=$(gomod-pseudo-version)
+            local cache_dir="${GOPATH}/pkg/mod/cache/download/${base_package}/${dep}/@v"
```
liggitt: I'd be hesitant to modify the modcache directory with our own "fake published" module versions. Would it be possible to do something like this instead:
1. run `go mod download` on k/k to fetch all non-staging dependencies into the module cache
2. copy content of `${GOPATH}/pkg/mod/cache/download` to `/some/goproxy/dir`
3. fake publish modules to `/some/goproxy/dir/${base_package}/${dep}/@v`
4. run module operations with `GOPROXY=file:///some/goproxy/dir`

artifacts/scripts/util.sh
```diff
@@ -787,180 +822,62 @@ function squash() {
...
+        GO111MODULE=on go mod edit -fmt -require "${base_package}/${dep}@${dep_commit}"
+        GO111MODULE=on go mod edit -fmt -replace "${base_package}/${dep}=${base_package}/${dep}@${dep_commit}"
     done
 
-    # due to a bug we have xxxx revisions for reflexive dependencies. Remove them.
-    jq '.Deps |= map(select(.ImportPath | (startswith("'${base_package}/$(basename "${PWD}")'/") or . == "'${base_package}/$(basename "${PWD}")'") | not))' Godeps/Godeps.json | indent-godeps > Godeps/Godeps.json.clean
-    mv Godeps/Godeps.json.clean Godeps/Godeps.json
-
-    git add Godeps/Godeps.json
+    GO111MODULE=on go mod tidy
```
liggitt: is GOPROXY set at this point?

configs/kubernetes-nightly-configmap.yaml
```diff
@@ -4,9 +4,10 @@ metadata:
   name: publisher-config
 data:
   config: |
-    source-org: kubernetes
+    source-org: liggitt
     source-repo: kubernetes
+    base-package: k8s.io
     target-org: kubernetes-nightly
-    rules-file: https://raw.githubusercontent.com/kubernetes/kubernetes/master/staging/publishing/rules.yaml
+    rules-file: https://raw.githubusercontent.com/sttts/kubernetes/sttts-go-mod-rules/staging/publishing/rules.yaml
```
liggitt: do I need to fold this into my PR?

liggitt: added smoke-test changes in https://github.com/kubernetes/kubernetes/pull/74877/commits/b3527f41f01f900954e378eb8d78a0b5a72dbfde

sttts:
/hold

until https://github.com/kubernetes/kubernetes/pull/74877 merges.

PR #177, opened 2019-05-07
Add docs about running publishing-bot in CNCF infra

nikhita:
For https://github.com/kubernetes/k8s.io/issues/183

The file should probably be renamed but I can't think of anything better/apt right now.

/kind documentation
/assign @dims @sttts


spiffxp:
/lgtm
Parts of this are going to be out of date as we move clusters / groups, but I'd rather merge now than wait

PR #183, opened 2019-06-10
Enable sync-tags only for main publishing-bot

nikhita:
sync-tags is [disabled](https://github.com/kubernetes/publishing-bot/commit/31e2757859d1f3d65a8593e502c5f0880ad105bc) for the main publishing-bot because gomodules
support hasn't been implemented yet. But it remained enabled for the
godeps branch, which meant that tagging still took place but didn't
include go.mod files at all.

This commit removes sync-tags support from the godeps branch, so that it
can be enabled for the main publishing-bot from the master branch later
(after gomodules support has been added).

Relevant slack thread - https://kubernetes.slack.com/archives/C2C40FMNF/p1560158651078000

/assign @sttts @dims

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -141,18 +141,4 @@ PUSH_SCRIPT=../push-tags-${REPO}-${DST_BRANCH}.sh
 echo "#!/bin/bash" > ${PUSH_SCRIPT}
 chmod +x ${PUSH_SCRIPT}
 
-if [ -z "${SKIP_TAGS}" ]; then
-    /sync-tags --prefix "$(echo ${SOURCE_REPO_NAME})-" \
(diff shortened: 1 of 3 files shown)


artifacts/scripts/construct.sh
```diff
@@ -141,18 +141,4 @@ PUSH_SCRIPT=../push-tags-${REPO}-${DST_BRANCH}.sh
 echo "#!/bin/bash" > ${PUSH_SCRIPT}
 chmod +x ${PUSH_SCRIPT}
 
-if [ -z "${SKIP_TAGS}" ]; then
```
sttts: why not just disable this block?

nikhita: Do you mean comment it out? I figured we would never want to run it from the godeps branch so we could just remove it.

PR #184, opened 2019-06-10
sync-tags: add gomodules support

nikhita:
This commit adds gomodules support for tagging (to sync-tags) to make sure
that kube dependencies point to the commit sha at the relevant tag in go.mod.

If the repo at the tag has a go.mod file, sync-tags pins the dependencies
at the pseudoversion pointing to the tag and packages up into the local
go mod cache, so that it can be downloaded and used by `go mod tidy`.
After that, godeps-gen is used to generate Godeps.json from the go.mod file.

If the repo at the tag does not contain a go.mod file, sync-tags directly
updates Godeps.json to point to the commit sha at the tag (old behaviour).

/cc @sttts @dims @liggitt 
/assign @sttts

### DIFF
--- a/cmd/sync-tags/gomod.go
+++ b/cmd/sync-tags/gomod.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// updateGomodWithTaggedDependencies gets the dependencies at the given tag and fills go.mod and go.sum.
+// If anything is changed, it commits the changes. Returns true if go.mod changed.
+func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, error) {
+	found := map[string]bool{}
+	changed := false
+
+	for _, dep := range depsRepo {
+		depPath := filepath.Join("..", dep)
+		dr, err := gogit.PlainOpen(depPath)
+		if err != nil {
+			return changed, fmt.Errorf("failed to open dependency repo at %q: %v", depPath, err)
+		}
+
+		depPkg, err := fullPackageName(depPath)
+		if err != nil {
+			return changed, fmt.Errorf("failed to get package at %s: %v", depPath, err)
+		}
+
+		commit, commitTime, err := taggedCommitHashAndTime(dr, tag)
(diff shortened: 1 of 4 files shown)


cmd/sync-tags/gomod.go
```diff
@@ -0,0 +1,69 @@
...
+			return changed, fmt.Errorf("failed to get package at %s: %v", depPath, err)
+		}
+
+		requireCommand := exec.Command("go", "mod", "edit", "-fmt", "-require", fmt.Sprintf("%s@%s", depPkg, tag))
+		requireCommand.Env = append(os.Environ(), "GO111MODULE=on")
+		if err := requireCommand.Run(); err != nil {
+			return changed, fmt.Errorf("Unable to pin %s in the require section of go.mod to %s: %v", depPkg, tag, err)
+		}
+
+		replaceCommand := exec.Command("go", "mod", "edit", "-fmt", "-replace", fmt.Sprintf("%s=%s@%s", depPkg, depPkg, tag))
```
liggitt: I don't think replace can use non-semantic, non-normalized tags

cmd/sync-tags/gomod.go
```diff
@@ -0,0 +1,69 @@
...
+	changed := false
+
+	for _, dep := range depsRepo {
+		depPath := filepath.Join("..", dep)
+		depPkg, err := fullPackageName(depPath)
+		if err != nil {
+			return changed, fmt.Errorf("failed to get package at %s: %v", depPath, err)
+		}
+
+		requireCommand := exec.Command("go", "mod", "edit", "-fmt", "-require", fmt.Sprintf("%s@%s", depPkg, tag))
```
liggitt: we shouldn't put non-semantic tags here that wouldn't survive a `go mod tidy`

nikhita:
Updated. PTAL.

nikhita:
Side note: interestingly, go allows to do a
`GO111MODULE=on go mod edit -fmt -require k8s.io/apimachinery@kubernetes-1.14.3`
but gives an error while doing
`GO111MODULE=on go mod edit -fmt -replace k8s.io/apimachinery=k8s.io/apimachinery@kubernetes-1.14.3`:

```
invalid module version "kubernetes-1.14.3": version must be of the form v1.2.3
```

sttts:
@nikhita have you smoke tested this and verified the created tags?

nikhita:
> @nikhita have you smoke tested this and verified the created tags?

Not sure how to do that right now, iiuc this would require a new tag being created in k/k to be able to test...unless the old incorrect tags are deleted or renamed.

sttts:
> unless the old incorrect tags are deleted or renamed.

Add some lines in the bot to delete them for testing.

nikhita:
/hold

This won't work with 1.12, 1.13, 1.14 tags since they'll need godeps.json generation in sync-tags directly (instead of generation from go.mod). will update.

nikhita:
Updated. PTAL.

Ran this locally for api and apimachinery:

**apimachinery**:
- `kubernetes-1.16.0-alpha.0` tag: https://github.com/prowtest/apimachinery/tree/kubernetes-1.16.0-alpha.0
- It doesn't have any dependencies on staging repos, so there is no extra commit here.

**api**:
- `kubernetes-1.16.0-alpha.0` tag: https://github.com/prowtest/api/tree/kubernetes-1.16.0-alpha.0
  - https://github.com/prowtest/api/commit/aa246005d2ed5d305125a35cb969071287033d74 is the extra commit fixing go.mod, go.sum and Godeps.json files.
- `kubernetes-1.14.0-alpha.3` tag: https://github.com/prowtest/api/tree/kubernetes-1.14.0-alpha.3
  - https://github.com/prowtest/api/commit/94f7f8e4bf97c5e58ce0462d33dde9be68a957b4 is the extra commit containing only Godeps.json changes because there is no go.mod in this case.

PR #185, opened 2019-06-18
Fix go mod publishing

nikhita:
This PR has the following changes:

- Display the error that might occur while running a `exec.Command`.
- Run `go mod download` before packaging up a dependency _and_ before updating the require and replace directives. This will help in avoiding packaging up, when the pseudo version is the same but we don't have it in the local cache.
- Fix the git command used for pseudo version calculation (#186)
- Publish all tags, not just alpha ones.
- Add a new go binary `gomod-zip` for creating the zip file that would have the same hash as that of the zip file if it were downloaded by `go mod download`. The zip file created by gomod differs from the former:
  - there are some extra normalizations: not considering anything in vendor, making sure we don't have directories in the file archive. These can, however, still be taken care of by `-x` and `-D` options. 
  - The zip file created by go mod download has the `00-00-1980 00:00` timestamp in the file header
for all files in the zip archive. This is not a valid UNIX timestamp and cannot be set easily. This is, however, valid in MSDOS. The `archive/zip` package [uses the MSDOS version](https://github.com/golang/go/blob/e9188d1d18040bc6cb46065b2474664b8728a6df/src/archive/zip/struct.go#L239-L243) so we create the zip file using this package.
    - change the initial zip file creation to use `git archive`: [go uses this](https://github.com/golang/go/blob/55d31e16c12c38d36811bdee65ac1f7772148250/src/cmd/go/internal/modfetch/codehost/git.go#L767).
    - add basic normalizations as [done by go](https://github.com/golang/go/blob/master/src/cmd/go/internal/modfetch/coderepo.go#L497) internally. 
    - use the `archive/zip` package to re-create the zip file from that created using `git archive`.

For the record: we had to do a `GO111MODULE=on go mod clean -modcache` first to clear the old zip files in our cache. Did this through `kubectl exec`, not using this PR.

For additional context, this [slack thread](https://kubernetes.slack.com/archives/C2C40FMNF/p1560775049352100) has more details.

/hold
doesn't fix the error yet, but I figured I'll open a PR to get my latest changes in

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ ADD _output/sync-tags /sync-tags
 ADD _output/init-repo /init-repo
 
 ADD _output/godeps-gen /godeps-gen
+ADD _output/gomod-zip /gomod-zip
 ADD artifacts/scripts/ /publish_scripts
 
 CMD ["/publishing-bot", "--dry-run", "--token-file=/token"]
--- a/cmd/gomod-zip/zip.go
+++ b/cmd/gomod-zip/zip.go
@@ -1,0 +1,216 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"archive/zip"
+	"bytes"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+const (
+	// MaxZipFile is the maximum size of downloaded zip file
+	MaxZipFile = 500 << 20
+)
+
+func Usage() {
+	fmt.Fprintf(os.Stderr, `Creates a zip file at
+$GOPATH/pkg/mod/cache/download/<package-name>/@v/<pseudo-version>.zip.
+The zip file has the same hash as if it were created by go mod download.
+This tool can be used to package modules which haven't been uploaded anywhere
+yet and are only available locally.
+
+This tool assumes that the package is already checked out at the commit
+pointed by the pseudo-version.
+
+package-name should be equal to the import path of the package.
+
+Usage: %s --package-name <package-name> --pseudo-version <pseudo-version>
+`, os.Args[0])
+	flag.PrintDefaults()
+}
+
+func main() {
+	packageName := flag.String("package-name", "", "package to zip")
+	pseudoVersion := flag.String("pseudo-version", "", "pseudoVersion to zip at")
+	flag.Parse()
+
+	if *packageName == "" {
+		glog.Fatalf("package-name cannot be empty")
+	}
+
+	if *pseudoVersion == "" {
+		glog.Fatalf("pseudo-version cannot be empty")
+	}
+
+	// create a zip file using git archive, and remove it after using it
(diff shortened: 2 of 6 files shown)


cmd/sync-tags/gomod.go
```diff
@@ -55,30 +55,46 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 		rev := commit.String()
 		pseudoVersion := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
 
+		downloadCommand := exec.Command("go", "mod", "download")
```
sttts: will this always pick the right go version?

cmd/sync-tags/gomod.go
```diff
@@ -55,30 +55,46 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
...
+		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on")
+		downloadCommand.Stdout = os.Stdout
+		downloadCommand.Stderr = os.Stderr
+		if err := downloadCommand.Run(); err != nil {
+			return changed, fmt.Errorf("Error running go mod download for %s: %v", depPkg, err)
```
sttts: lower case errors

cmd/sync-tags/gomod.go
nikhita: Not always, but it _could_ if the pseudo version doesn't change. Added a comment.

Dockerfile
```diff
@@ -37,6 +37,7 @@ ADD _output/sync-tags /sync-tags
 ADD _output/init-repo /init-repo
 
 ADD _output/godeps-gen /godeps-gen
+ADD _output/godeps-gen /gomod-zip
```
sttts: typo?

cmd/gomod-zip/zip.go
```diff
@@ -0,0 +1,215 @@
...
+	dl := ioutil.NopCloser(bytes.NewReader(archive))
+	defer dl.Close()
+
+	// Spool to local file.
+	f, err := ioutil.TempFile("", "gomodzip-")
+	if err != nil {
+		dl.Close()
+		glog.Fatalf("error creating temp file: %v", err)
+	}
+	defer os.Remove(f.Name())
```
sttts: does Name include the path?

cmd/gomod-zip/zip.go
```diff
@@ -0,0 +1,215 @@
...
+	defer f.Close()
+
+	maxSize := int64(MaxZipFile)
+	lr := &io.LimitedReader{R: dl, N: maxSize + 1}
+	if _, err := io.Copy(f, lr); err != nil {
+		dl.Close()
+		glog.Fatalf("error reading from %s: %v", f.Name(), err)
+	}
+	dl.Close()
```
sttts: there is a defer Close as well

cmd/gomod-zip/zip.go
```diff
@@ -0,0 +1,215 @@
...
+
+	packagedZipPath := fmt.Sprintf("%s/pkg/mod/cache/download/%s/@v/%s.zip", os.Getenv("GOPATH"), *packageName, *pseudoVersion)
+	dst, err := os.OpenFile(packagedZipPath, os.O_CREATE|os.O_WRONLY, 0755)
+	if err != nil {
+		glog.Fatalf("failed to create zip file at %s: %v", packagedZipPath, err)
+	}
+	defer dst.Close()
+	zw := zip.NewWriter(dst)
+
+	for _, zf := range zr.File {
```
sttts: write where this is from

cmd/gomod-zip/zip.go
nikhita: Yes

Dockerfile
nikhita: yikes, fixed

cmd/sync-tags/main.go
```diff
@@ -193,11 +193,6 @@ func main() {
 			continue
 		}
 
-		// temporarily publish only alpha tags
-		if !strings.Contains(bName, "alpha") {
```
nikhita: Removed the alpha tags restriction. PTAL.

PR #188, opened 2019-06-24
Cope with undefined PUBLISHER_BOT_GENERATE_GODEPS

sttts:
It's not set if skipGodeps is false in the rules file.


nikhita:
/lgtm

thanks!

PR #190, opened 2019-06-25
Fix sync-tags --generate-godeps syntax error

sttts:


sttts:
@nikhita OpenShift runs 4 bots without godep/gomod generation.

PR #196, opened 2019-07-07
[WIP] sync-tags: add support for versioned imports

nikhita:
This PR adds support for tagging a version which contains versioned imports. For `v1.15.1` of k/k, two versions will be published:
- `kubernetes-1.15.1` - this tag will _not_ contain versioned imports, and will instead rely on pseudo versions (existing behaviour).
- `v15.0.1` - this new tag will contain versioned imports.

To update import paths, this uses https://github.com/marwan-at-work/mod. `mod` is also "officially" suggested in the [go modules wiki](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher). We use the `v0.2.2` version of this tool because we need https://github.com/marwan-at-work/mod/pull/8 for it to work.

Overall, we do the following:

- Update the import paths containing the current module and update the module name in `go.mod`.
- If the repo has dependencies, we then do the following for each dependency:
  - Fake-publish/package the dependency at the new tag to the local go mod cache.
  - Update all import paths containing this dependency.
  - Add require and replace directives for the new tag, and `dropreplace` the old pseudoversion.
  - Run `go mod tidy`. This drops the old require directives and updates `go.sum`.

An example run of this can be seen at [apimachinery@v15.0.0](https://github.com/prowtest/apimachinery/tree/v15.0.0), [api@v15.0.0](https://github.com/prowtest/api/tree/v15.0.0), [client-go@v15.0.0](https://github.com/prowtest/client-go/tree/v15.0.0).

/cc @sttts @liggitt @dims 
/assign @sttts

### DIFF
--- a/cmd/sync-tags/gomod.go
+++ b/cmd/sync-tags/gomod.go
@@ -58,7 +58,7 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 		// in case the pseudoVersion has not changed, running go mod download will help
 		// in avoiding packaging it up if the pseudoVersion has been published already
 		downloadCommand := exec.Command("go", "mod", "download")
-		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on")
+		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on", "GOPOXY=file://${GOPATH}/pkg/mod/cache/download")
 		downloadCommand.Stdout = os.Stdout
 		downloadCommand.Stderr = os.Stderr
 		if err := downloadCommand.Run(); err != nil {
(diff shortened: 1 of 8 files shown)


nikhita:
This PR (almost) generates the desired output but still has a few problems. Thought I'd open the PR to get some initial feedback though.

- Running `GO111MODULE=on go build ./...` on the new tag for client-go leads to errors like `tools/reference/ref.go:113:3: cannot use objectMeta.GetUID() (type "k8s.io/apimachinery/v15/pkg/types".UID) as type "k8s.io/apimachinery/pkg/types".UID in field value`.

- Somehow _after_ tagging client-go, we are _always_ left with a dirty git worktree and the only changes are that `rest/request_test.go` and `tools/cache/store.go` files have executable permissions (?!?!?). I haven't been able to come with a reasonable conclusion why this error occurs and why it's only these two files every time yet.

- The `replace` directives don't seem to be formatted correctly even though the `-fmt` flag is used: https://github.com/prowtest/client-go/blob/v15.0.0/go.mod#L42.

- Updating import paths also updates the paths in proto files, not sure if that's ok?

/hold

cmd/sync-tags/gomod.go
```diff
@@ -58,7 +58,7 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 		// in case the pseudoVersion has not changed, running go mod download will help
 		// in avoiding packaging it up if the pseudoVersion has been published already
 		downloadCommand := exec.Command("go", "mod", "download")
-		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on")
+		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on", "GOPOXY=file://${GOPATH}/pkg/mod/cache/download")
```
liggitt: typo: GOPROXY

liggitt: (GOPOXY is only useful for glue code)

nikhita:
Quick update here: I'm looking into using https://godoc.org/golang.org/x/tools/go/packages for gengo this week, since code generation doesn't seem to like versioned imports. :/

Not entirely sure if that'll solve the problem, but want to experiment first. Ref: https://golang.org/doc/go1.11#gopackages, https://github.com/kubernetes/gengo/issues/147

```
➜  sample-controller git:(538b921d) ✗ GO111MODULE=on ./hack/update-codegen.sh 
can't load package: package k8s.io/code-generator/v15/cmd/defaulter-gen: unknown import path "k8s.io/code-generator/v15/cmd/defaulter-gen": cannot find package
can't load package: package k8s.io/code-generator/v15/cmd/client-gen: unknown import path "k8s.io/code-generator/v15/cmd/client-gen": cannot find package
can't load package: package k8s.io/code-generator/v15/cmd/lister-gen: unknown import path "k8s.io/code-generator/v15/cmd/lister-gen": cannot find package
can't load package: package k8s.io/code-generator/v15/cmd/informer-gen: unknown import path "k8s.io/code-generator/v15/cmd/informer-gen": cannot find package
can't load package: package k8s.io/code-generator/v15/cmd/deepcopy-gen: unknown import path "k8s.io/code-generator/v15/cmd/deepcopy-gen": cannot find package
```

liggitt:
not sure how much of this is for go1.13 and how much is for module mode, but some random hacks I did to get k/k building with go1.13 are in https://github.com/liggitt/kubernetes/commits/go-1.13 in case it helps

nikhita:
Going to close this until we decide we want something like this + hash out the design/make changes to the KEP for the same.

/close

PR #202, opened 2019-08-16
Bump default Go version to 1.12.9

tao12345666333:
Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

ref: https://github.com/kubernetes/kubernetes/pull/81489


nikhita:
/hold cancel

/assign @sttts 
for approval

tao12345666333:
> Looks like we should move this to the rules file

I think I missed this reply. :joy_cat:  What do you mean?

PR #203, opened 2019-09-12
Add merge commit to be picked as single commit

nikhita:
https://github.com/kubernetes/kubernetes/commit/64718f678695884c93d6d3df8f5799614746bea2 is the merge commit.

This is done because https://github.com/kubernetes/kubernetes/pull/81453 has
changes split across multiple commits and leads to conflicts while linearizing
history. This, however, doesn't occur while picking the merge commit as a single commit
because the problematic changes are nullified across the commits in the PR.

/assign @sttts


nikhita:
Redeployed.

PR #204, opened 2019-10-14
Use proxy.golang.org for go mod download

nikhita:
Ref: https://github.com/kubernetes/kubernetes/issues/83761

If a dependency is re-tagged after the dependency is updated in k/k, the
hash in go.sum in k/k will point to the old tag.

When we run `go mod download` in the publishing-bot, the bot currently
downloads directly from github.com. Since the dependency was re-tagged,
the downloaded hash points to the new tag and is no longer equal to
the old hash in go.sum. This leads to a "checksum mismatch error".

Reverting the bad dependency update in k/k won't help with the
publishing-bot because we check if go.mod needs update after each commit.

To fix the bot while encountering such "bad" changes, this commit uses
proxy.golang.org as GOPROXY while running `go mod download` since
proxy.golang.org caches content and the old hash still exists there.

Note: proxy.golang.org is also used in the CI for k/k, which ensures
that the verify job remains green in case of such re-tagging.

/cc @sttts @dims @liggitt 
/assign @sttts

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -857,7 +857,7 @@ update-deps-in-gomod() {
 
     GO111MODULE=on go mod edit -json | jq -r '.Replace[]? | select(.New.Path | startswith("../")) | "-dropreplace \(.Old.Path)"' | GO111MODULE=on xargs -L 100 go mod edit -fmt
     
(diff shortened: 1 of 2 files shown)


nikhita:
Also verified locally.

nikhita:
Redeployed.

PR #206, opened 2019-11-12
Set GOPRIVATE to avoid validating unpublished SHAs

nikhita:
From Go 1.13, the go command defaults to downloading modules from the
public Go module mirror at proxy.golang.org. It also defaults to
validating downloaded modules, regardless of source, against the public
Go checksum database at sum.golang.org.

To avoid this validation for all non-published SHAs of staging repos,
we set the GOPRIVATE env var.

Ref: https://golang.org/cmd/go/#hdr-Module_configuration_for_non_public_modules

Without this, we get errors like:

```
+ GO111MODULE=on
+ GOPROXY=https://proxy.golang.org
+ go mod download
verifying k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: reading https://sum.golang.org/lookup/k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da: 410 Gone
verifying k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: reading https://sum.golang.org/lookup/k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da: 410 Gone
verifying k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da/go.mod: reading https://sum.golang.org/lookup/k8s.io/api@v0.0.0-20191112131459-ab1b4ad002da: 410 Gone
```

/assign @sttts @liggitt @dims

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -847,6 +847,12 @@ update-deps-in-gomod() {
     local dep_count=${#deps_array[@]}
     local base_package=${2}
 
+    # if dependencies exist, dep_packages is a comma separated list of {base_package}/{dep}. Eg: "k8s.io/api,k8s.io/apimachinery"
+    local dep_packages=""
+    if [ "$dep_count" != 0 ]; then
+      dep_packages="$(echo ${1} | tr "," "\n" | sed -e 's/:.*//' -e s/^/"${base_package}\/"/ | paste -sd "," -)"
+    fi
+
     for (( i=0; i<${dep_count}; i++ )); do
         local dep="${deps_array[i]%%:*}"
         local dep_commit=$(cd ../${dep}; gomod-pseudo-version)
@@ -857,7 +863,7 @@ update-deps-in-gomod() {
 
     GO111MODULE=on go mod edit -json | jq -r '.Replace[]? | select(.New.Path | startswith("../")) | "-dropreplace \(.Old.Path)"' | GO111MODULE=on xargs -L 100 go mod edit -fmt
     
-    GO111MODULE=on GOPROXY=https://proxy.golang.org go mod download
+    GO111MODULE=on GOPRIVATE="${dep_packages}" GOPROXY=https://proxy.golang.org go mod download
     GOPROXY="file://${GOPATH}/pkg/mod/cache/download" GO111MODULE=on go mod tidy
 
     git add go.mod go.sum
@@ -890,6 +896,13 @@ checkout-deps-to-kube-commit() {
     local deps=()
     IFS=',' read -a deps <<< "${2}"
     local base_package=${3}
+    local dep_count=${#deps[@]}
+
+    # if dependencies exist, dep_packages is a comma separated list of {base_package}/{dep}. Eg: "k8s.io/api,k8s.io/apimachinery"
+    local dep_packages=""
+    if [ "$dep_count" != 0 ]; then
+      dep_packages="$(echo ${2} | tr "," "\n" | sed -e 's/:.*//' -e s/^/"${base_package}\/"/ | paste -sd "," -)"
+    fi
 
     # get last k8s.io/kubernetes commit on HEAD ...
     local k_last_kube_commit="$(last-kube-commit ${commit_msg_tag} HEAD)"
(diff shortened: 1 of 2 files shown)


nikhita:
https://github.com/kubernetes/publishing-bot/pull/205 and https://github.com/kubernetes/kubernetes/pull/85088 moved the publishing-bot to use 1.13. I redeployed the bot after both PRs merged and the bot broke...

artifacts/scripts/util.sh
```diff
@@ -921,7 +921,7 @@ checkout-deps-to-kube-commit() {
             git checkout -q "${dep_commit}"
 
             echo "Downloading go mod dependencies..."
-            GO111MODULE=on GOPROXY=https://proxy.golang.org go mod download
+            GO111MODULE=on GOPRIVATE="k8s.io/*" GOPROXY=https://proxy.golang.org go mod download
```
sttts: this should come from the config, or be the source repo org, not hardcoded.

nikhita: Ah, right. Will update.

cmd/sync-tags/main.go
```diff
@@ -73,6 +73,7 @@ func main() {
 	skipFetch := flag.Bool("skip-fetch", false, "skip fetching tags")
 	generateGodeps := flag.Bool("generate-godeps", false, "regenerate Godeps.json from go.mod")
 	mappingOutputFile := flag.String("mapping-output-file", "", "a file name to write the source->dest hash mapping to ({{.Tag}} is substituted with the tag name, {{.Branch}} with the local branch name)")
+	basePackage := flag.String("base-pkg", "k8s.io", "the canonical package name (eg. k8s.io)")
```
nikhita: FYI this is a "breaking change" for non-k8s repos, but they'll need this if they decide to use go 1.13 anyway.

We also default `base_package` to k8s.io in `util.sh` today, unless specified.

liggitt: I don't object to requiring the base package, but I wouldn't expect this to default to k8s.io. I'd expect invokers to be required to provide it now

artifacts/scripts/util.sh
```diff
@@ -921,7 +921,7 @@ checkout-deps-to-kube-commit() {
             git checkout -q "${dep_commit}"
 
             echo "Downloading go mod dependencies..."
-            GO111MODULE=on GOPROXY=https://proxy.golang.org go mod download
+            GO111MODULE=on GOPRIVATE="${base_package}/*" GOPROXY=https://proxy.golang.org go mod download
```
liggitt: will this break dependencies on things like k8s.io/utils?

nikhita: I think it won't break, but it'll skip validating the checksums for it.

I think this approach would be better - we have all dependencies (repo+branch name) in `deps`. We can then use `base_package` + repo names to define GOPRIVATE as an explicit list of all staging repo dependencies.

/hold
I'll address the reviews and verify that this works as expected locally

liggitt: > I think this approach would be better - we have all dependencies (repo+branch name) in `deps`. We can then use `base_package` + repo names to define GOPRIVATE as an explicit list of all staging repo dependencies.

That does sound better. Targeted and handled internally

cmd/sync-tags/main.go
sttts: I thought we have this concept already in some form. Target org or so?

nikhita: > I thought we have this concept already in some form. Target org or so?


We don't have any info about the base package name in sync-tags, we "derive" it from the package path:

https://github.com/kubernetes/publishing-bot/blob/ec388bdeb34233252eb18babc440d82a385c91b8/cmd/sync-tags/gomod.go#L40

I updated to use the path to avoid adding another flag. PTAL.

nikhita:
Addressed comments and tested locally.

PR #207, opened 2019-11-13
Drop unused replace directives

### LINKED ISSUE #214: Remove no-op replace directives from go.mod files
If the require and replace directive for a dependency match after running `go mod tidy`, drop the replace directive, since it is a no-op and confuses consumers

/assign

liggitt:
Tested locally, but would be good to run in nightly.

cc @sttts @nikhita 

Fixes #214

xref https://github.com/kubernetes/kubernetes/issues/85105

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -860,6 +860,18 @@ update-deps-in-gomod() {
     GO111MODULE=on GOPROXY=https://proxy.golang.org go mod download
     GOPROXY="file://${GOPATH}/pkg/mod/cache/download" GO111MODULE=on go mod tidy
 
+    # After tidying, drop any replace directives that match the naturally preferred require directive.
+    # 1. List the common lines where replace and require directives match exactly
+    comm -12 \
+      <(GO111MODULE=on go mod edit -json | jq -r '.Replace // [] | .[] | "\(.New.Path) \(.New.Version)"') \
+      <(GO111MODULE=on go mod edit -json | jq -r '.Require // [] | .[] | "\(.Path) \(.Version)"') \
+    | \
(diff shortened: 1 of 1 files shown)


nikhita:
While I agree that this is safer, can we directly `dropreplace` the dependencies?

This would also need to be updated in https://github.com/kubernetes/publishing-bot/blob/master/cmd/sync-tags/gomod.go.

liggitt:
> While I agree that this is safer, can we directly `dropreplace` the dependencies?

I'm not sure what you're asking

liggitt:
is this the relevant section:
https://github.com/kubernetes/publishing-bot/blob/master/cmd/sync-tags/gomod.go#L82-L88

I think it might be better to avoid adding the replace there in the first place if it doesn't already exist

nikhita:
Would it be possible to skip the step in line 855 where we set the replace directive in the first place, instead of removing it after adding it? And then use `dropreplace` to replace the original replace directives like `k8s.io/api => ../api`?

https://github.com/kubernetes/publishing-bot/blob/ec388bdeb34233252eb18babc440d82a385c91b8/artifacts/scripts/util.sh#L854-L858

nikhita:
> is this the relevant section:
https://github.com/kubernetes/publishing-bot/blob/master/cmd/sync-tags/gomod.go#L82-L88

yes

> I think it might be better to avoid adding the replace there in the first place if it doesn't already exist

I was wondering if something similar was also possible for util.sh as well.. :)

liggitt:
I don't want to drop the replace until `go mod tidy` resolves/verifies it matches our naturally preferred version

liggitt:
reopened in https://github.com/kubernetes/publishing-bot/pull/415 with updated version

PR #209, opened 2019-11-27
Add merge commit to be picked as single commit

nikhita:
https://github.com/kubernetes/kubernetes/commit/bc53b97ceb25338570a853845c4cdd295468ed61 is the merge commit.

This is done because https://github.com/kubernetes/kubernetes/pull/85464 has
changes split across multiple commits and leads to conflicts while linearizing
history. This, however, doesn't occur while picking the merge commit as a single commit
because the problematic changes are nullified across the commits in the PR.

/assign @sttts 
cc @dims


nikhita:
This does not seem to work. I'm travelling tomorrow, but will take a look later.

PR #210, opened 2019-12-04
sync-tags: add support for v0.x.y tags

nikhita:
For https://github.com/kubernetes/kubernetes/issues/84608

Implements changes to the go modules KEP made in
https://github.com/kubernetes/enhancements/pull/1350.

This commit updates sync-tags to publish `v0.x.y` tags for corresponding
`v1.x.y` Kubernetes tag. The old `kubernetes-1.x.y` tags are still also
published. Both `kubernetes-1.x.y` and `v0.x.y` point to the same sha.

For `v0.x.y`:

- At this tag, `go.mod` uses `v0.x.y` tags for dependencies.
- When this tag is manually specified in a `go.mod` file, it will be
retained as `v0.x.y`.

For `kubernetes-1.x.y`:

- At this tag, `go.mod` uses `v0.x.y` tags for dependencies. Note: this
differs from the previous way, where pseudoversions were used.
- When this tag is manually specified in a `go.mod` file, go will resolve
it to the appropriate pseudoversion.
---

Verified locally. A sample run for client-go (and api + apimachinery) is shown below:

- [v0.17.0](https://github.com/prowtest/client-go/tree/v0.17.0) = [kubernetes-1.17.0](https://github.com/prowtest/client-go/tree/kubernetes-1.17.0)
- [v0.17.0-rc.2](https://github.com/prowtest/client-go/tree/v0.17.0-rc.2) = [kubernetes-1.17.0-rc.2](https://github.com/prowtest/client-go/tree/kubernetes-1.17.0-rc.2)
- [v.17.0-beta.2](https://github.com/prowtest/client-go/tree/v0.17.0-beta.2)  = [kubernetes-1.17.0-beta.2](https://github.com/prowtest/client-go/tree/kubernetes-1.17.0-beta.2) 
- [v0.18.0-alpha.0](https://github.com/prowtest/client-go/tree/v0.18.0-alpha.0)  = [kubernetes-1.18.0-alpha.0](https://github.com/prowtest/client-go/tree/kubernetes-1.18.0-alpha.0) 

/cc @sttts @liggitt @dims 
/assign @sttts

### DIFF
--- a/cmd/sync-tags/gomod.go
+++ b/cmd/sync-tags/gomod.go
@@ -33,7 +33,7 @@ import (
 
 // updateGomodWithTaggedDependencies gets the dependencies at the given tag and fills go.mod and go.sum.
 // If anything is changed, it commits the changes. Returns true if go.mod changed.
-func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, error) {
+func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag bool) (bool, error) {
 	found := map[string]bool{}
 	changed := false
 
@@ -59,10 +59,14 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 			return changed, fmt.Errorf("failed to get tag %s for %q: %v", tag, depPkg, err)
 		}
 		rev := commit.String()
-		pseudoVersion := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
+		pseudoVersionOrTag := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
 
-		// in case the pseudoVersion has not changed, running go mod download will help
-		// in avoiding packaging it up if the pseudoVersion has been published already
+		if semverTag {
+			pseudoVersionOrTag = tag
+		}
+
+		// in case the pseudoVersion/tag has not changed, running go mod download will help
+		// in avoiding packaging it up if the pseudoVersion/tag has been published already
 		downloadCommand := exec.Command("go", "mod", "download")
 		downloadCommand.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPRIVATE=%s", depPackages), "GOPROXY=https://proxy.golang.org")
 		downloadCommand.Stdout = os.Stdout
@@ -71,26 +75,26 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 			return changed, fmt.Errorf("error running go mod download for %s: %v", depPkg, err)
 		}
 
-		// check if we have the pseudoVersion published already. if we don't, package it up
+		// check if we have the pseudoVersion/tag published already. if we don't, package it up
 		// and save to local mod download cache.
-		if err := packageDepToGoModCache(depPath, depPkg, rev, pseudoVersion, commitTime); err != nil {
+		if err := packageDepToGoModCache(depPath, depPkg, rev, pseudoVersionOrTag, commitTime); err != nil {
 			return changed, fmt.Errorf("failed to package %s dependency: %v", depPkg, err)
 		}
 
-		requireCommand := exec.Command("go", "mod", "edit", "-fmt", "-require", fmt.Sprintf("%s@%s", depPkg, pseudoVersion))
+		requireCommand := exec.Command("go", "mod", "edit", "-fmt", "-require", fmt.Sprintf("%s@%s", depPkg, pseudoVersionOrTag))
 		requireCommand.Env = append(os.Environ(), "GO111MODULE=on")
 		requireCommand.Stdout = os.Stdout
 		requireCommand.Stderr = os.Stderr
 		if err := requireCommand.Run(); err != nil {
-			return changed, fmt.Errorf("unable to pin %s in the require section of go.mod to %s: %v", depPkg, pseudoVersion, err)
+			return changed, fmt.Errorf("unable to pin %s in the require section of go.mod to %s: %v", depPkg, pseudoVersionOrTag, err)
 		}
 
-		replaceCommand := exec.Command("go", "mod", "edit", "-fmt", "-replace", fmt.Sprintf("%s=%s@%s", depPkg, depPkg, pseudoVersion))
+		replaceCommand := exec.Command("go", "mod", "edit", "-fmt", "-replace", fmt.Sprintf("%s=%s@%s", depPkg, depPkg, pseudoVersionOrTag))
 		replaceCommand.Env = append(os.Environ(), "GO111MODULE=on")
 		replaceCommand.Stdout = os.Stdout
 		replaceCommand.Stderr = os.Stderr
 		if err := replaceCommand.Run(); err != nil {
-			return changed, fmt.Errorf("unable to pin %s in the replace section of go.mod to %s: %v", depPkg, pseudoVersion, err)
+			return changed, fmt.Errorf("unable to pin %s in the replace section of go.mod to %s: %v", depPkg, pseudoVersionOrTag, err)
 		}
 
 		downloadCommand2 := exec.Command("go", "mod", "download")
@@ -98,11 +102,11 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 		downloadCommand2.Stdout = os.Stdout
 		downloadCommand2.Stderr = os.Stderr
 		if err := downloadCommand2.Run(); err != nil {
-			return changed, fmt.Errorf("error running go mod download for pseudo-version %s for %s: %v", pseudoVersion, depPkg, err)
+			return changed, fmt.Errorf("error running go mod download for pseudo-version %s for %s: %v", pseudoVersionOrTag, depPkg, err)
 		}
 
 		tidyCommand := exec.Command("go", "mod", "tidy")
-		tidyCommand.Env = append(os.Environ(), "GO111MODULE=on", "GOPOXY=file://${GOPATH}/pkg/mod/cache/download")
+		tidyCommand.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPROXY=file://%s/pkg/mod/cache/download", os.Getenv("GOPATH")))
 		tidyCommand.Stdout = os.Stdout
 		tidyCommand.Stderr = os.Stderr
 		if err := tidyCommand.Run(); err != nil {
@@ -110,7 +114,7 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 		}
 
 		found[dep] = true
-		fmt.Printf("Bumping %s in go.mod to %s\n.", depPkg, rev)
+		fmt.Printf("Bumping %s in go.mod to %s.\n", depPkg, rev)
 		changed = true
 	}
 
@@ -146,18 +150,18 @@ type ModuleInfo struct {
 	Time    string
 }
 
-func packageDepToGoModCache(depPath, depPkg, commit, pseudoVersion string, commitTime time.Time) error {
+func packageDepToGoModCache(depPath, depPkg, commit, pseudoVersionOrTag string, commitTime time.Time) error {
 	cacheDir := fmt.Sprintf("%s/pkg/mod/cache/download/%s/@v", os.Getenv("GOPATH"), depPkg)
-	goModFile := fmt.Sprintf("%s/%s.mod", cacheDir, pseudoVersion)
+	goModFile := fmt.Sprintf("%s/%s.mod", cacheDir, pseudoVersionOrTag)
 
 	if _, err := os.Stat(goModFile); err == nil {
-		fmt.Printf("Pseudo version %s for %s is already packaged up.\n", pseudoVersion, depPkg)
+		fmt.Printf("%s for %s is already packaged up.\n", pseudoVersionOrTag, depPkg)
 		return nil
 	} else if err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("Could not check if %s exists: %v", goModFile, err)
 	}
 
-	fmt.Printf("Packaging up pseudo version %s for %s into go mod cache.\n", pseudoVersion, depPkg)
+	fmt.Printf("Packaging up %s for %s into go mod cache.\n", pseudoVersionOrTag, depPkg)
 
 	// create the cache if it doesn't exist
 	if err := os.MkdirAll(filepath.Dir(goModFile), os.FileMode(755)); err != nil {
@@ -173,14 +177,14 @@ func packageDepToGoModCache(depPath, depPkg, commit, pseudoVersion string, commi
 		return fmt.Errorf("failed to checkout %s at %s: %v", depPkg, commit, err)
 	}
 
-	// copy go.mod to pseudoVersion.mod in the cache dir
+	// copy go.mod to the cache dir
 	if err := copyFile(fmt.Sprintf("%s/go.mod", depPath), goModFile); err != nil {
 		return fmt.Errorf("unable to copy %s file to %s to gomod cache for %s: %v", fmt.Sprintf("%s/go.mod", depPath), goModFile, depPkg, err)
 	}
 
-	// create pseudoVersion.info file in the cache dir
+	// create info file in the cache dir
 	moduleInfo := ModuleInfo{
-		Version: pseudoVersion,
+		Version: pseudoVersionOrTag,
 		Name:    commit,
 		Short:   commit[:12],
 		Time:    commitTime.UTC().Format("2006-01-02T15:04:05Z"),
@@ -190,17 +194,17 @@ func packageDepToGoModCache(depPath, depPkg, commit, pseudoVersion string, commi
 	if err != nil {
 		return fmt.Errorf("error marshaling .info file for %s: %v", depPkg, err)
 	}
-	if err := ioutil.WriteFile(fmt.Sprintf("%s/%s.info", cacheDir, pseudoVersion), moduleFile, 0644); err != nil {
-		return fmt.Errorf("failed to write %s file for %s: %v", fmt.Sprintf("%s/%s.info", cacheDir, pseudoVersion), depPkg, err)
+	if err := ioutil.WriteFile(fmt.Sprintf("%s/%s.info", cacheDir, pseudoVersionOrTag), moduleFile, 0644); err != nil {
+		return fmt.Errorf("failed to write %s file for %s: %v", fmt.Sprintf("%s/%s.info", cacheDir, pseudoVersionOrTag), depPkg, err)
 	}
 
-	// create the pseudoVersion.zip file in the cache dir. This zip file has the same hash
+	// create the zip file in the cache dir. This zip file has the same hash
 	// as of the zip file that would have been created by go mod download.
-	zipCommand := exec.Command("/gomod-zip", "--package-name", depPkg, "--pseudo-version", pseudoVersion)
+	zipCommand := exec.Command("/gomod-zip", "--package-name", depPkg, "--pseudo-version", pseudoVersionOrTag)
 	zipCommand.Stdout = os.Stdout
 	zipCommand.Stderr = os.Stderr
 	if err := zipCommand.Run(); err != nil {
-		return fmt.Errorf("failed to run gomod-zip for %s at %s: %v", depPkg, pseudoVersion, err)
+		return fmt.Errorf("failed to run gomod-zip for %s at %s: %v", depPkg, pseudoVersionOrTag, err)
 	}
 
 	// append the pseudoVersion to the list file in the cache dir
@@ -210,7 +214,7 @@ func packageDepToGoModCache(depPath, depPkg, commit, pseudoVersion string, commi
 	}
 	defer listFile.Close()
 
-	if _, err := listFile.WriteString(fmt.Sprintf("%s\n", pseudoVersion)); err != nil {
+	if _, err := listFile.WriteString(fmt.Sprintf("%s\n", pseudoVersionOrTag)); err != nil {
 		return fmt.Errorf("unable to write to list file in %s: %v", cacheDir, err)
 	}
 
--- a/cmd/sync-tags/main.go
+++ b/cmd/sync-tags/main.go
@@ -28,6 +28,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/golang/glog"
 	"github.com/renstrom/dedent"
 	gogit "gopkg.in/src-d/go-git.v4"
@@ -73,6 +74,7 @@ func main() {
 	skipFetch := flag.Bool("skip-fetch", false, "skip fetching tags")
 	generateGodeps := flag.Bool("generate-godeps", false, "regenerate Godeps.json from go.mod")
 	mappingOutputFile := flag.String("mapping-output-file", "", "a file name to write the source->dest hash mapping to ({{.Tag}} is substituted with the tag name, {{.Branch}} with the local branch name)")
+	publishSemverTags := flag.Bool("publish-v0-semver", false, "publish v0.x.y tag at destination repo for v1.x.y tag at the source repo")
 
 	flag.Usage = Usage
 	flag.Parse()
@@ -187,6 +189,19 @@ func main() {
 			bName = *prefix + name[1:] // remove the v
 		}
 
+		var (
+			semverTag        = ""
+			publishSemverTag = false
+		)
+		// if we are publishing semver tags
+		if *publishSemverTags {
+			// and this is a valid v1... semver tag
+			if _, semverErr := semver.Parse(name[1:]); semverErr == nil && strings.HasPrefix(name, "v1.") {
+				publishSemverTag = true
+				semverTag = "v0." + strings.TrimPrefix(name, "v1.") // replace v1.x.y with v0.x.y
+			}
+		}
+
 		// ignore non-annotated tags
 		tag, err := r.TagObject(kh)
 		if err != nil {
@@ -199,14 +214,41 @@ func main() {
 			continue
 		}
 
-		// skip if it already exists in origin
-		if _, found := bTagCommits[bName]; found {
+		// skip if either tag exists at origin
+		_, nonSemverTagAtOrigin := bTagCommits[bName]
+		_, semverTagAtOrigin := bTagCommits[semverTag]
+		if nonSemverTagAtOrigin || (publishSemverTag && semverTagAtOrigin) {
 			continue
 		}
 
-		// do not override tags (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
+		// if any of the tag exists locally,
+		// delete the tags, clear the cache and recreate them
 		if tagExists(r, bName) {
-			continue
+			commit, commitTime, err := taggedCommitHashAndTime(r, bName)
+			if err != nil {
+				glog.Fatalf("Failed to get tag %s: %v", bName, err)
+			}
+			rev := commit.String()
+			pseudoVersion := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
+
+			fmt.Printf("Clearing cache for local tag %s.\n", pseudoVersion)
+			if err := cleanCacheForTag(pseudoVersion); err != nil {
+				glog.Fatalf("Failed to clean go mod cache for %s: %v", pseudoVersion, err)
+			}
+
+			if err := deleteTag(bName); err != nil {
+				glog.Fatalf("Failed to delete tag %s: %v", bName, err)
+			}
+		}
+
+		if publishSemverTag && tagExists(r, semverTag) {
+			fmt.Printf("Clearing cache for local tag %s.\n", semverTag)
+			if err := cleanCacheForTag(semverTag); err != nil {
+				glog.Fatalf("Failed to clean go mod cache for %s: %v", semverTag, err)
+			}
+			if err := deleteTag(semverTag); err != nil {
+				glog.Fatalf("Failed to delete tag %s: %v", semverTag, err)
+			}
 		}
 
 		// lazily compute kube commit map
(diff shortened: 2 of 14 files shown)


cmd/sync-tags/gomod.go
```diff
@@ -59,10 +59,14 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
 			return changed, fmt.Errorf("failed to get tag %s for %q: %v", tag, depPkg, err)
 		}
 		rev := commit.String()
-		pseudoVersion := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
+		pseudoVersionOrTag := fmt.Sprintf("v0.0.0-%s-%s", commitTime.UTC().Format("20060102150405"), rev[:12])
```
liggitt: as soon as we start adding tags, v0.0.0 will no longer be the correct pseudoversion for commits in between semver tags... see https://tip.golang.org/cmd/go/#hdr-Pseudo_versions

My reading of that means that when generating a non-tagged version:
* if the nearest semver tag is `v0.17.0`, we should generate a pseudoversion using the format string `v0.17.1-0.%s-%s`
* if the nearest semver tag is `v1.17.0-alpha.3`, we should generate a pseudoversion using the format string `v1.17.0-alpha.3.0.%s-%s`

sttts: isn‘t there automatic normalization through go mod we can use?

liggitt: that means we need to know the nearest existing tag for the current repo... not sure whether we've fetched that info and if it is available currently

liggitt: > isn‘t there automatic normalization through go mod we can use?

I'm not sure how much of that is available to us before the commits/tags are pushed to the authoritative VCS

nikhita: It isn't available currently, but once we fetch that looks like we can plug it into https://github.com/golang/vgo/blob/master/vendor/cmd/go/internal/modfetch/pseudo.go#L48

cmd/sync-tags/main.go
```diff
@@ -323,6 +306,31 @@ func main() {
 			glog.Fatalf("Failed to create tag %q: %v", bName, err)
 		}
 		createdTags = append(createdTags, bName)
+
+		// if publishSemverTags is true, create annotated semver tag
+		if *publishSemverTags {
```
liggitt: does this script sync *all* existing tags? will this try to create semver tags corresponding to all the existing kubernetes release tags?

cmd/sync-tags/main.go
```diff
@@ -323,6 +306,31 @@ func main() {
...
+			semverTag := name[:1] + "0" + name[2:] // replace v1.x.y with v0.x.y
```
liggitt: this mapping of upstream v1.x.y upstream -> v0.x.y staging is kube-specific... most of the other options in this script try to remain generic. @sttts might have thoughts on this

cmd/sync-tags/main.go
```diff
@@ -323,6 +306,31 @@ func main() {
...
+
+			if goModChanged && len(dependentRepos) > 0 {
+				wt := checkoutBranchTagCommit(r, bh, dependentRepos)
+				changed = updateGoModAndGodeps(semverTag, dependentRepos, *generateGodeps, true)
```
liggitt: by creating the prefixed tag matching upstream (kubernetes-1.x.y) first, and the semver tag second, isn't this no longer re-entrant if the script fails during the second tag creation (since we return early if bTagCommits[bName] already exists)?

liggitt: also, won't calling updateGoModAndGodeps again mean that we'll change the go.mod files before adding this tag, so `kubernetes-1.x.y` and `v0.x.y` tags will not actually point to the same sha? I expected both tags to point to the same content

nikhita: >  isn't this no longer re-entrant if the script fails during the second tag creation (since we return early if bTagCommits[bName] already exists)?

will fix

> also, won't calling updateGoModAndGodeps again mean that we'll change the go.mod files before adding this tag, so kubernetes-1.x.y and v0.x.y tags will not actually point to the same sha? I expected both tags to point to the same content

My understanding is that we'd want to use `v0.x.y` tags in go.mod while publishing the `v0.x.y` tags (example - https://github.com/prowtest/client-go/blob/v0.17.0-rc.2/go.mod#L41-L42).

To keep the sha same for both `kubernetes-1.x.y` and `v0.x.y` (makes sense), should we then also use `v0.x.y` tags in go.mod while publishing `kubernetes-1.x.y` i.e. same go.mod content for `kubernetes-1.17.0-rc.2` as the link above?

cmd/sync-tags/main.go
nikhita: No, only newer tags but will make this simpler while fixing https://github.com/kubernetes/publishing-bot/pull/210#discussion_r354082521

cmd/sync-tags/main.go
liggitt: > To keep the sha same for both kubernetes-1.x.y and v0.x.y (makes sense), should we then also use v0.x.y tags in go.mod while publishing kubernetes-1.x.y i.e. same go.mod content for kubernetes-1.17.0-rc.2 as the link above and we won't use pseudoversions in tags anymore?

That was my expectation

cmd/sync-tags/gomod.go
nikhita: Updated to not use pseudoversions anymore, so this problem is fixed.

cmd/sync-tags/main.go
nikhita: Updated.

cmd/sync-tags/main.go
nikhita: Updated.

nikhita:
Addressed all comments except https://github.com/kubernetes/publishing-bot/pull/210#discussion_r353966692. Also, updated the main PR body and commit messages to reflect the new behaviour and added links to a new test run in the PR body.

If `--publish-semver-tags` is true, we won't use pseudoversions, so the problem described in https://github.com/kubernetes/publishing-bot/pull/210#discussion_r353916045 is not relevant anymore.

cmd/sync-tags/main.go
```diff
@@ -199,15 +206,20 @@ func main() {
...
+		} else if nonSemverTagExists {
 			continue
 		}
 
 		// do not override tags (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
-		if tagExists(r, bName) {
-			continue
-		}
+		// if tagExists(r, bName) {
+		//	continue
```
nikhita: @sttts question about this if statement: why is this required? don't we do the same check in:

https://github.com/kubernetes/publishing-bot/blob/dc65f29519695e122f8f4aa7dc1269ab6997884a/cmd/sync-tags/main.go#L202-L205

cmd/sync-tags/gomod.go
sttts: but tags? why didn't we come up with that in the first place? :)

cmd/sync-tags/gomod.go
```diff
@@ -71,46 +75,46 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string) (bool, err
...
 		downloadCommand2.Stdout = os.Stdout
 		downloadCommand2.Stderr = os.Stderr
 		if err := downloadCommand2.Run(); err != nil {
-			return changed, fmt.Errorf("error running go mod download for pseudo-version %s for %s: %v", pseudoVersion, depPkg, err)
+			return changed, fmt.Errorf("error running go mod download for pseudo-version %s for %s: %v", pseudoVersionOrTag, depPkg, err)
 		}
 
 		tidyCommand := exec.Command("go", "mod", "tidy")
-		tidyCommand.Env = append(os.Environ(), "GO111MODULE=on", "GOPOXY=file://${GOPATH}/pkg/mod/cache/download")
+		tidyCommand.Env = append(os.Environ(), "GO111MODULE=on")
```
sttts: why is this not necessary anymore?

cmd/sync-tags/main.go
sttts: wasn't the bName check above for the origin tag only? I think we rewrite tag names on fetch to `origin/<name>` or something like that. So we have local tags and origin tags. This check is for local tags, the upper for origin.

cmd/sync-tags/main.go
```diff
@@ -258,71 +270,69 @@ func main() {
...
+			wt := checkoutBranchTagCommit(r, bh, dependentRepos)
 
 			// if go.mod exists, fix only go.mod and generate Godeps.json from it later
 			// if it doesn't exist, check if Godeps.json exists, and update it
-			var changed, goModChanged bool
-			_, err = os.Stat("go.mod")
-			if os.IsNotExist(err) {
+			if !goModExists {
 				if _, err2 := os.Stat("Godeps/Godeps.json"); err2 == nil {
 					fmt.Printf("Updating Godeps.json to point to %s tag.\n", bName)
```
sttts: should we replace Godeps.json with go.mod everywhere?

sttts: or is this really still about godeps?

cmd/sync-tags/main.go
nikhita: Ah right, will update.

cmd/sync-tags/gomod.go
nikhita: GOPROXY is not needed while running go mod tidy. Moreover, the original code had a typo "GOPOXY", so it was never really used. :)

cmd/sync-tags/main.go
nikhita: This is for the case where there is no go.mod but only Godeps.json file in the repo. We don't need it for k8s repos but might be useful for non-k8s ones.

Or we could remove it here and force them to use the `godeps` branch in this repo, but that'd break compatibility.

cmd/sync-tags/main.go
```diff
@@ -199,13 +206,19 @@ func main() {
...
+		semverTagExists := semverTagAtOrigin || tagExists(r, semverTag)
 
-		// do not override tags (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
-		if tagExists(r, bName) {
+		// skip tags if they exist locally or at origin already
+		// we want to ensure to not override tags
+		// (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
+		if *publishSemverTags && semverTagExists && nonSemverTagExists {
+			continue
+		} else if nonSemverTagExists {
```
nikhita: Updated the check for tags at origin and locally. Verified this and pushed the latest results to the test repo, but can you sanity check this again?

cmd/sync-tags/gomod.go
liggitt: > GOPROXY is not needed while running go mod tidy

really? tidy requires resolving dependencies, and this module has dependencies on unpublished tags/shas... I don't understand how this is working without GOPROXY pointing at our cache

cmd/sync-tags/main.go
liggitt: does this script still run on 1.14.x, or did we split the job to use the godeps branch for publishing that branch? dropping godeps manipulation seems ok, but I'd do it in a separate PR

cmd/sync-tags/main.go
```diff
@@ -488,3 +499,48 @@ func regenerateGodepsFromGoMod() error {
...
+		}
+	}
+	return changed
+}
+
+func createCommitToFixDeps(wt *gogit.Worktree, tag string) plumbing.Hash {
+	fmt.Printf("Adding extra commit fixing dependencies to point to %s tags.\n", tag)
+	publishingBotNow := publishingBot
+	publishingBotNow.When = time.Now()
+	bh, err := wt.Commit(fmt.Sprintf("Fix dependencies to point to %s tag", tag), &gogit.CommitOptions{
```
liggitt: nit: prefer "Update dependencies to %s tag"

cmd/sync-tags/gomod.go
liggitt: are we inheriting a GOPROXY setting from our parent env?

liggitt:
The tags published to https://github.com/prowtest look good

cmd/sync-tags/gomod.go
nikhita: > are we inheriting a GOPROXY setting from our parent env?

No, I don't think so. I tried running this locally as well and it did not globally try setting GOPROXY fwiw.

I _think_ `go mod tidy` looks at the local mod cache before donwloading from vcs...

Running the following commands at client-go checked out at `v0.18.0-alpha.0`:

- With the local mod cache cleared:

```
➜  client-go git:(2768422194b) GO111MODULE=on go mod tidy -v                                                                                                                                               0s 
get "k8s.io/apimachinery": found meta tag get.metaImport{Prefix:"k8s.io/apimachinery", VCS:"git", RepoRoot:"https://github.com/kubernetes/apimachinery"} at //k8s.io/apimachinery?go-get=1
go: k8s.io/apimachinery@v0.18.0-alpha.0: reading k8s.io/apimachinery/go.mod at revision v0.18.0-alpha.0: unknown revision v0.18.0-alpha.0
```

- After adding a `v0.18.0-alpha.0.mod` at `${GOPATH}/pkg/mod/cache/download/k8s.io/apimachinery/@v/`, `GO111MODULE=on go mod tidy -v` does not produce any output/errors.

cmd/sync-tags/main.go
```diff
@@ -463,7 +474,7 @@ func mappingOutputFileName(fnameTpl string, branch, tag string) string {
 
 func regenerateGodepsFromGoMod() error {
 	goListCommand := exec.Command("go", "list", "-m", "-json", "all")
-	goListCommand.Env = append(os.Environ(), "GO111MODULE=on", "GOPOXY=file://${GOPATH}/pkg/mod/cache/download")
+	goListCommand.Env = append(os.Environ(), "GO111MODULE=on")
```
nikhita: Found another GOPOXY btw :/

cmd/sync-tags/main.go
nikhita: > does this script still run on 1.14.x, or did we split the job to use the godeps branch for publishing that branch?

The latter. Will remove this in another PR. :+1:

cmd/sync-tags/main.go
```diff
@@ -182,10 +183,16 @@ func main() {
...
+		// temporarily publish only alpha tags
+		if !strings.Contains(name, "alpha") {
+			continue
+		}
+
 		bName := name
 		if *prefix != "" {
 			bName = *prefix + name[1:] // remove the v
 		}
+		semverTag := name[:1] + "0" + name[2:] // replace v1.x.y with v0.x.y
```
liggitt: at the very last, only do this swap to replace `v1.` with `v0.` (otherwise, we'd collapse v1.x.y and v2.x.y to the same tags)

```
semverTag = name
if strings.HasPrefix(name, "v1.") {
  semverTag = "v0." + strings.TrimPrefix(name, "v1.")
}
```

liggitt: also, make sure name is actually a valid semver tag before triggering the semver path

nikhita: Done. Also updated the results of the latest run to https://github.com/prowtest. PTAL.

cmd/sync-tags/gomod.go
liggitt: > With the local mod cache cleared:
> ...
> go: k8s.io/apimachinery@v0.18.0-alpha.0: reading k8s.io/apimachinery/go.mod at revision v0.18.0-alpha.0: unknown revision v0.18.0-alpha.0

did that fail the tidy command? did that fail our sync-tags command? (it should)

> After adding a v0.18.0-alpha.0.mod at ${GOPATH}/pkg/mod/cache/download/k8s.io/apimachinery/@v/, GO111MODULE=on go mod tidy -v does not produce any output/errors.

are we currently creating the `v0.18.0-alpha.0.mod` file?

nikhita: > did that fail the tidy command?

When I manually cleared my local cache and ran `go mod tidy`, yes.

> did that fail our sync-tags command? (it should)
> are we currently creating the v0.18.0-alpha.0.mod file?

We publish the `.mod`, `.info`, `.zip` files for each tag at the local cache, so we have a `v0.18.0-alpha.0.mod` file. And since we have this file, the sync-tags command does not fail.

https://github.com/kubernetes/publishing-bot/blob/3f3695e10c68b22812723076d843d2e48a65feaf/cmd/sync-tags/gomod.go#L80

liggitt: ok, so it sounds like we're getting the implicit behavior of "check the cache dir, then hit the network", and we happen to have the things we want in the cache dir.

I still think we should formalize that by pointing GOPROXY at the cache dir (fixing the typo in the envvar) rather than removing it. If the item is not in the cache dir, I really don't want us talking to the network to try to find things.

nikhita: > I still think we should formalize that by pointing GOPROXY at the cache dir (fixing the typo in the envvar, rather than removing it).

Done.

cmd/sync-tags/main.go
nikhita: > this mapping of upstream v1.x.y upstream -> v0.x.y staging is kube-specific... most of the other options in this script try to remain generic. 

For the record: we discussed this on slack and decided to go with renaming the flag to `--publish-v0-semver` to represent the v1 -> v0 semver mapping.

cmd/sync-tags/main.go
```diff
@@ -182,11 +184,29 @@ func main() {
 	// create or update tags from kTagCommits as local tags with the given prefix
 	createdTags := []string{}
 	for name, kh := range kTagCommits {
+		// temporarily publish only alpha tags
+		if !strings.Contains(name, "alpha") {
```
liggitt: Was this present when you ran in the prow test org? If so, how did the rc.2 tag get published?

nikhita: Not present, and I manually deleted the rc2, beta2 and alpha0 tags in published repos to trigger runs for those. If not, sync-tags wouldn't sync any tags (it syncs only newer tags).

cmd/sync-tags/main.go
```diff
@@ -182,11 +184,29 @@ func main() {
...
+		}
+
 		bName := name
 		if *prefix != "" {
 			bName = *prefix + name[1:] // remove the v
 		}
 
+		// check if name is a semver tag
+		// the semver lib requires the v prefix to be dropped before parsing
+		var skipPublishingSemverTag bool
```
liggitt: considering both publishSemverTags and skipPublishingSemverTag in many places is confusing. suggest something more like:

```go
var (
  semverTag = ""
  publishSemverTag = false
)
// if we are publishing semver tags
if *publishSemverTags {
  // and this is a valid v1... semver tag
  if _, semverErr := semver.Parse(name[1:]); semverErr == nil && strings.HasPrefix(name, "v1.") {
    publishSemverTag = true
    semverTag = "v0." + strings.TrimPrefix(name, "v1.")
  }
}
```

and then only consider the `publishSemverTag` boolean in the following logic

cmd/sync-tags/main.go
liggitt: I would drop this, and just do dry runs against the test org

nikhita: > I would drop this, and just do dry runs against the test org

If we drop this and merge this PR before Monday, `kubernetes-1.17.0` and `v0.17.0` would be automatically published once `v1.17.0` is tagged in k/k... I don't think that would give us enough time to do a run of `v1.17.0` on a test org.

liggitt: would it make sense to:
1. separately merge a commit that skips syncing tags
2. let k8s cut v1.17.0
3. make this branch re-enable syncing tags, and do test runs against a test org
4. merge this and re-enable tag sync once ready?

that would allow us to get this in for 1.17.0 without rushing

nikhita: Sounds good. :+1: 

Will make the changes in a bit.

cmd/sync-tags/main.go
nikhita: Done (and tested).

cmd/sync-tags/main.go
nikhita: Dropped this change and created https://github.com/kubernetes/kubernetes/pull/86008 to disable syncing tags.

cmd/sync-tags/main.go
```diff
@@ -199,13 +214,19 @@ func main() {
...
-		// do not override tags (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
-		if tagExists(r, bName) {
+		// skip tags if they exist locally or at origin already
+		// we want to ensure to not override tags
+		// (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
+		if publishSemverTag && semverTagExists && nonSemverTagExists {
+			continue
+		} else if nonSemverTagExists {
 			continue
 		}
```
liggitt: this bit still looks strange to me. if either tag exists at origin, we cannot make changes and should skip, right? something like this, maybe?

```go
if publishSemverTag && semverTagAtOrigin {
  continue
}
if nonSemverTagAtOrigin {
  continue
}
```

for the local checks, I'm not sure how re-entrant this script is expected to be (e.g. if it crashes after creating a local non-semver tag, but before pushing, do we wipe/reclone on next run?)

liggitt:
Overall approach and results lgtm, I just had the one comment about re-entrancy and origin tags already existing. Will defer to @sttts on that, as he is much more familiar with the flow this is executed/retried in

cmd/sync-tags/main.go
nikhita: Will let @sttts confirm but this is my understanding:

> if it crashes after creating a local non-semver tag, but before pushing, do we wipe/reclone on next run?

The tag will be [appended in the push script](https://github.com/kubernetes/publishing-bot/blob/23fba6b1430ce9a03fc61bde591c4e1bdbf7eb84/cmd/sync-tags/main.go#L349) once it's created, so it'll try pushing again when the push script is executed on the next run.

>  if either tag exists at origin, we cannot make changes and should skip, right?

Because the script handles cases when tags have been created locally but not pushed, I think we'll need to consider cases when the tag exists locally as well (and not just whether it exists at origin).

> if either tag exists

I thought of skipping tag creation if _one_ of them exists but then considered the following cases...though I'm not entirely sure when we'd hit these cases. Maybe if someone deletes a tag? However, retagging on tag deletion is also a bad idea... :thinking: 

- If `kubernetes-1.x.y` exists but `v0.x.y` does not, the go.mod file for `kubernetes-1.x.y` would point to `v0.x.y` so we **need** to make sure that `v0.x.y` is published.

- If `v0.x.y` exists but `kubernetes-1.x.y` does not, we publish it because both go hand-in-hand (though not publishing it doesn't harm in any way).

sttts: > The tag will be appended in the push script once it's created, so it'll try pushing again when the push script is executed on the next run.

If I remember right, the push script is recreated on each run (I hope so). Also local tags should be remove before each run.

The mixed logic for semver and non-semver irritates me. I would make all code below idem-potent. Then we could skip this early continue here, but we do it for purely performance reason to skip the git checkout magic if there is no new tag.

Then what @nikhita has in the code looks right. Could be rewritten to the following:

```
if nonSemverTagExists && (!publishSemverTag || semverTagExists) {
  continue
}
```

cmd/sync-tags/main.go
```diff
@@ -258,71 +279,69 @@ func main() {
...
-
-			if goModChanged && *generateGodeps {
-				fmt.Printf("Regenerating Godeps.json from go.mod.\n")
-				if err := regenerateGodepsFromGoMod(); err != nil {
-					glog.Fatalf("Failed to regenerate Godeps.json from go.mod: %v", err)
+			} else {
+				if publishSemverTag {
+					changed = updateGoModAndGodeps(semverTag, dependentRepos, *generateGodeps, true)
+				} else {
+					changed = updateGoModAndGodeps(bName, dependentRepos, *generateGodeps, false)
```
sttts: does this mean we don't support semver tags and `kubernetes-*` tags in parallel?

cmd/sync-tags/main.go
```diff
@@ -258,71 +279,69 @@ func main() {
...
+			fmt.Printf("Tagging %v as %q.\n", bh, bName)
+			err = createAnnotatedTag(bh, bName, tag.Tagger.When, dedent.Dedent(fmt.Sprintf(`
+				Kubernetes release %s
+
+				Based on https://github.com/kubernetes/kubernetes/releases/tag/%s
+				`, name, name)))
+			if err != nil {
+				glog.Fatalf("Failed to create tag %q: %v", bName, err)
+			}
+			createdTags = append(createdTags, bName)
```
sttts: all this look idem-potent and independent. What confuses me is that the commits these tags (semver and non-semver) use, are the same. I would expect different ones, pointing to the respective tags of the dependencies.

nikhita: > What confuses me is that the commits these tags (semver and non-semver) use, are the same. I would expect different ones, pointing to the respective tags of the dependencies.

That was the initial approach, but that exposed us to the pseudoversion problem. Having them point to the `v0.x.y` is not ideal, but that solves the pseudoversion problem for us.

Also, having them point to different commits would mean that `kubernetes-1.x.y` and `v0.x.y` won't be equivalent (in terms of sha) anymore. Prior discussion: https://github.com/kubernetes/publishing-bot/pull/210#discussion_r354098096

cmd/sync-tags/main.go
nikhita: > does this mean we don't support semver tags and kubernetes-* tags in parallel?

We do. Both of them point to the same sha, so we create tags at the sha later (like you commented below).

sttts: annotated tags lead to different shas. Doesn't this mean conflicting shas if I pin all dependencies to kubernetes-v1.17.0, but e.g. client-go kuberentes-v1.17.0 points to apimachinery v0.17.0 which is different from apimachiner kubernetes-v1.17.0 ?

nikhita:
Just adding an explicit hold until we've tested this thoroughly
/hold

cmd/sync-tags/main.go
liggitt: > annotated tags lead to different shas

really? https://github.com/prowtest/client-go/releases shows identical shas for kubernetes-1.17.0-rc.2 and v0.17.0-rc.2

cmd/sync-tags/main.go
liggitt: > What confuses me is that the commits these tags (semver and non-semver) use, are the same. I would expect different ones, pointing to the respective tags of the dependencies.

There is little benefit to shipping a non-semver tag with a go.mod pointing to other non-semver tags, since the tags are tidied to v0.0.0-... before commit, making the go.mod files almost unusable.

nikhita:
Updated with the latest changes, PTAL. The current workflow (as discussed on slack) is:

1) figure out all the tags we want to create
2) if any already exist remotely, we're done
3) if any already exist locally, delete them (and also delete the corresponding go mod cache .mod/.zip files)
4) update+tidy+commit the go.mod (with the semver tag if we're semver-tagging, with the pseudoversion otherwise), create all tags, then push all tags with --atomic

Note: for 3) we don't overwrite the existing .mod/.info/.zip files in the go mod cache dir while building tags instead because with the current logic, we avoid creating the `.mod`, etc files if they already exist.

https://github.com/kubernetes/publishing-bot/blob/dde80ae17679962763a799410d6e54e0eb6df361/cmd/sync-tags/gomod.go#L157-L159

Even if we tried to keep track of overwriting it _once_, it isn't possible to do it across multiple runs (for multiple repos) of sync-tags.

cmd/sync-tags/main.go
```diff
@@ -199,14 +214,35 @@ func main() {
...
 
-		// do not override tags (we build master first, i.e. the x.y.z-alpha.0 tag on master will not be created for feature branches)
-		if tagExists(r, bName) {
-			continue
+		// if any of the tag exists locally,
+		// delete the tags, clear the cache and recreate them
+		if tagExists(r, bName) || (publishSemverTag && tagExists(r, semverTag)) {
+			if err := deleteTag(bName); err != nil {
+				glog.Fatalf("Failed to delete tag %s: %v", bName, err)
+			}
```
liggitt: won't this fail if the condition that got us into this block was `publishSemverTag && tagExists(r, semverTag)`?

liggitt: I expected something more like:
```
if tagExists(r, bName) {
  if err := deleteTag(bName); err != nil {
     glog.Fatalf("Failed to delete tag %s: %v", bName, err)
  }
  if err := cleanCacheForTag(bName); err != nil {
    glog.Fatalf("Failed to clean go mod cache for %s: %v", bName, err)
  }
}
if publishSemverTag && tagExists(r, semverTag) {
  if err := deleteTag(semverTag); err != nil {
     glog.Fatalf("Failed to delete tag %s: %v", semverTag, err)
  }
  if err := cleanCacheForTag(semverTag); err != nil {
    glog.Fatalf("Failed to clean go mod cache for %s: %v", semverTag, err)
  }
}
```

liggitt: it also looks like we should call cleanCacheForTag first, otherwise we won't enter this block if called again after a failure while calling cleanCacheForTag (the git tag will be gone and we won't enter this block)

nikhita: Addressed these comments. PTAL.

I think there were two more things to consider:

- We can't call `cleanCacheForTag(bName)` because the cache would contain pseudoversion.mod, pseudoversion.zip files, not bName.mod, etc files. We'd need to use `cleanCacheForTag(pseudoversion)`.

- We can't call `cleanCacheForTag(pseudoversion)` if bName exists -- we'd also need to check if we are publishing semver tags. In case of semver tagging, we'll have semverTag.mod files in the cache, instead of pseudoversion.mod files.

liggitt: > We can't call cleanCacheForTag(bName) because the cache would contain pseudoversion.mod, pseudoversion.zip files

Good point

> We can't call cleanCacheForTag(pseudoversion) just if bName exists -- we'd also need to check if we are publishing semver tags

cleanCacheForTag (correctly) tolerates absence of any of the files it cares about, so we should be able to call `cleanCacheForTag(pseudoversion)` unconditionally, right?

nikhita: > cleanCacheForTag (correctly) tolerates absence of any of the files it cares about, so we should be able to call cleanCacheForTag(pseudoversion) unconditionally, right?

done

nikhita:
Rebased on top of https://github.com/kubernetes/publishing-bot/pull/211

liggitt:
test run to kubernetes-nightly org looks good
* all repos got v0.17.0 and kubernetes-1.17.0 tags on the same sha
* all go.mod files referenced v0.17.0 tags for peer k8s.io dependencies

PR #211, opened 2019-12-10
Use gcr.io/k8s-staging-publishing-bot for nightly

nikhita:
If we use `sttts/k8s-publishing-bot-nightly`, @sttts would need to push the image everytime.

/hold
I'm not sure if we want to have the nightly version in the "official" gcr staging repo, so adding an explicit hold for this. If we don't want to have it there, we can always specify a personal docker repo while deploying the nightly bot too.

/assign @dims @sttts 
fyi @liggitt

### DIFF
--- a/configs/kubernetes-nightly
+++ b/configs/kubernetes-nightly
@@ -1,4 +1,4 @@
-DOCKER_REPO = sttts/k8s-publishing-bot-nightly
+DOCKER_REPO = gcr.io/k8s-staging-publishing-bot/k8s-publishing-bot-nightly
 NAMESPACE = k8s-publishing-bot-nightly
 SCHEDULE = * */4 * * *
 INTERVAL = 14400


liggitt:
yeah, I'm not sure about using gcr.io for this. @kubernetes/sig-testing do we have a image repository that would good to use for testing images?

dims:
discussion on slack, let's do this.

https://kubernetes.slack.com/archives/C2C40FMNF/p1575986202353000?thread_ts=1575978808.340300&cid=C2C40FMNF

/approve
/lgtm
/hold cancel

configs/kubernetes-nightly
```diff
@@ -1,4 +1,4 @@
-DOCKER_REPO = sttts/k8s-publishing-bot-nightly
```
liggitt: does the kubernetes-nightly-token file need to change as well?

PR #212, opened 2019-12-10
Update fetch and push commands in helper script

nikhita:
`git fetch --tags` is needed to fetch all tags and `git push --atomic`
will help in ensuring that we don't end up selectively pushing tags.

/cc @dims @liggitt @sttts

### DIFF
--- a/hack/fetch-all-latest-and-push.sh
+++ b/hack/fetch-all-latest-and-push.sh
@@ -88,18 +88,18 @@ for (( i=0; i<${repo_count}; i++ )); do
     branches=$(git branch -r | grep "^ *origin" | sed 's,^ *origin/,,' | grep -v HEAD | grep -v '^master' || true)
     tags=$(git tag | sed 's,^,refs/tags/,')
     if [ -n "${branches}${tags}" ]; then
-        git push --delete origin ${branches} ${tags}
+        git push --atomic --delete origin ${branches} ${tags}
     fi
 
     # push all upstream tags and branches to origin
     git tag | xargs git tag -d
-    git fetch upstream --prune -q
+    git fetch upstream --tags --prune -q
     branches=$(git branch -r | grep "^ *upstream" | sed 's,^ *upstream/,,' | grep -v HEAD || true)
     branches_arg=""
     for branch in ${branches}; do
         branches_arg+=" upstream/${branch}:refs/heads/${branch}"
     done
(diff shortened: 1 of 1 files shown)


hack/fetch-all-latest-and-push.sh
```diff
@@ -88,18 +88,18 @@ for (( i=0; i<${repo_count}; i++ )); do
...
     tags=$(git tag | sed 's,^,refs/tags/,')
     if [ -n "${branches}${tags}" ]; then
-        git push --delete origin ${branches} ${tags}
+        git push --atomic --delete origin ${branches} ${tags}
     fi
 
     # push all upstream tags and branches to origin
     git tag | xargs git tag -d
-    git fetch upstream --prune -q
+    git fetch --tags upstream --prune -q
```
nikhita: will fix the order

nikhita:
/assign @sttts 
for approval

hack/fetch-all-latest-and-push.sh
```diff
@@ -88,18 +88,18 @@ for (( i=0; i<${repo_count}; i++ )); do
...
     tags=$(git tag | sed 's,^,refs/tags/,')
     if [ -n "${branches}${tags}" ]; then
-        git push --delete origin ${branches} ${tags}
+        git push --atomic --delete origin ${branches} ${tags}
     fi
 
     # push all upstream tags and branches to origin
     git tag | xargs git tag -d
-    git fetch upstream --prune -q
+    git fetch upstream --tags --prune -q
```
sttts: why this change?

liggitt: without this, only tags on fetched branches are pulled. starting in 1.15, there was always an additional commit to fixup go.mod files, so the tags were never on the branches. without this, no tags >= 1.15.x were copied into the destination repos

liggitt: From git help:

> By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in. This default behavior can be changed by using the --tags or --no-tags options or by configuring remote.<name>.tagOpt. By using a refspec that fetches tags explicitly, you can fetch tags that do not point into branches you are interested in as well.

sttts: thanks. Makes sense.

Issue #217, opened 2020-02-20
F0220 16:09:13.608470 1882464 main.go:551] Failed to update go.mod and go.sum for tag v0.15.11-beta.0: failed to get tag v0.15.11-beta.0 for "k8s.io/api": failed to get refs/tags/v0.15.11-beta.0: reference not found


liggitt:
Persistent error in https://github.com/kubernetes/kubernetes/issues/56876#issuecomment-589158190

The failure message comes from here:

https://github.com/kubernetes/publishing-bot/blob/da9508d9d841255b64fc7ace6dc6010c5e225a5b/cmd/sync-tags/main.go#L217-L229

We should not get to that point if the semver tag already exists at the origin, [which it does](https://github.com/kubernetes/api/releases/tag/v0.15.11-beta.0)

/assign @nikhita @sttts

PR #218, opened 2020-02-21
sync-tags: on volume recreation tags might only exist as origin/<tag>

sttts:
Fixes https://github.com/kubernetes/publishing-bot/issues/217.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 FROM debian:stretch
 MAINTAINER Stefan Schimanski <sttts@redhat.com>
 RUN apt-get update \
- && apt-get install -y -qq git=1:2.11.0-3+deb9u4 \
+ && apt-get install -y -qq git=1:2.11.0-3+deb9u5 \
  && apt-get install -y -qq mercurial \
  && apt-get install -y -qq ca-certificates curl wget jq vim tmux bsdmainutils tig gcc zip \
  && rm -rf /var/lib/apt/lists/*
--- a/cmd/sync-tags/gomod.go
+++ b/cmd/sync-tags/gomod.go
@@ -54,7 +54,7 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag 
 			return changed, fmt.Errorf("failed to get package at %s: %v", depPath, err)
 		}
 
-		commit, commitTime, err := taggedCommitHashAndTime(dr, tag)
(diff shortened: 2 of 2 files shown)


sttts:
/hold

until proven to help.

Dockerfile
```diff
@@ -15,7 +15,7 @@
 FROM debian:stretch
 MAINTAINER Stefan Schimanski <sttts@redhat.com>
 RUN apt-get update \
- && apt-get install -y -qq git=1:2.11.0-3+deb9u4 \
+ && apt-get install -y -qq git=1:2.11.0-3+deb9u5 \
```
liggitt: ?

cmd/sync-tags/gomod.go
```diff
@@ -55,6 +55,10 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag
 		}
 
 		commit, commitTime, err := taggedCommitHashAndTime(dr, tag)
+		if err != nil {
+			// if tag has not been created on this volume, but pre-exists in Github, we have renamed that as origin/<tag>
+			commit, commitTime, err = taggedCommitHashAndTime(dr, "origin/"+tag)
```
liggitt: do we not get tags from origin because of this:

https://github.com/kubernetes/publishing-bot/blob/da9508d9d841255b64fc7ace6dc6010c5e225a5b/artifacts/scripts/construct.sh#L96-L97

I see a similar fallback in the godeps file, and slightly prefer a dedicated named function to do the fallback:

https://github.com/kubernetes/publishing-bot/blob/da9508d9d841255b64fc7ace6dc6010c5e225a5b/cmd/sync-tags/godeps.go#L132-L138

Dockerfile
sttts: Debian updated git and old one is gone. We have this explicit version here to notice when git changes.

cmd/sync-tags/gomod.go
sttts: ` git fetch origin --no-tags --prune` does not download tags. We have another place (have to look for it) and prefix them with `origin/` in order to control which are new and which are not.

liggitt: wrap this fallback into a `localOrPublishedTaggedCommitHashAndTime` method, then lgtm

sttts: done

PR #221, opened 2020-04-01
Prepare publishing-bot migration

ameukam:
Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,14 @@ TOKEN ?=
 KUBECTL ?= kubectl
 SCHEDULE ?= 0 5 * * *
 INTERVAL ?= 86400
+CPU_LIMITS ?= 2
+CPU_REQUESTS ?= 300m
 MEMORY_REQUESTS ?= 200Mi
 MEMORY_LIMITS ?= 1.6Gi
 GOOS ?= linux
 
 build_cmd = GO111MODULE=on mkdir -p _output && GOOS=$(GOOS) go build -o _output/$(1) ./cmd/$(1)
-prepare_spec = sed 's,DOCKER_IMAGE,$(DOCKER_REPO),g;s,MEMORY_REQUESTS,$(MEMORY_REQUESTS),g;s,MEMORY_LIMITS,$(MEMORY_LIMITS),g'
+prepare_spec = sed 's,DOCKER_IMAGE,$(DOCKER_REPO),g;s,CPU_LIMITS,$(CPU_LIMITS),g;s,CPU_REQUESTS,$(CPU_REQUESTS),g;s,MEMORY_REQUESTS,$(MEMORY_REQUESTS),g;s,MEMORY_LIMITS,$(MEMORY_LIMITS),g'
 
 SHELL := /bin/bash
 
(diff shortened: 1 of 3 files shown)


artifacts/manifests/cronjob.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
   name: publisher
+  namespace: publishing-bot
```
mfojtik, line 5: why is setting the namespace necessary?

ameukam: It's better to regroup all the resources related to publishing-bot in the same namespace instead of using the `ðefault` namespace on the new cluster.

mfojtik: i thought that if you don't specify the namespace `kubectl apply` will just use the "current" namespace.

ameukam: @mfojtik What's the issue by specifying the namespace ?

Makefile
```diff
@@ -24,7 +24,7 @@ TOKEN ?=
 KUBECTL ?= kubectl
 SCHEDULE ?= 0 5 * * *
 INTERVAL ?= 86400
-MEMORY_REQUESTS ?= 200Mi
+MEMORY_REQUESTS ?= 1.6Gi
```
sttts, line 27: I am against this default. It breaks other consumers. This variable is there in order to be customized. Feel free to do so from outside.

configs/kubernetes
```diff
@@ -1,6 +1,6 @@
 DOCKER_REPO = gcr.io/k8s-staging-publishing-bot/k8s-publishing-bot
-NAMESPACE = k8s-publishing-bot
+NAMESPACE = publishing-bot
 SCHEDULE = * */4 * * *
 INTERVAL = 14400
 MEMORY_REQUESTS = 2Gi
-MEMORY_LIMITS = 3Gi
+MEMORY_LIMITS = 2Gi
```
sttts, line 6: these changes seem to be legit if that's what kube needs for the deployment cc @nikhita

artifacts/manifests/cronjob.yaml
sttts: this code is generic. There are other consumers than just kube. Please don't change this. Namespace is intensionally missing.

artifacts/manifests/podspec.yaml
```diff
@@ -13,7 +13,6 @@ initContainers:
       cpu: 300m
       memory: 200Mi
     limits:
```
sttts, line 15: don't remove cpu limits please. Make them configurable like memory.

ameukam:
@sttts @nikhita About Tim's [suggestion](https://github.com/kubernetes/k8s.io/pull/520#discussion_r379147033), WDYT ?

artifacts/manifests/podspec.yaml
ameukam: It's done.

Issue #215, opened 2019-12-11
Consider forcing a diff in go.mod when syncing semver tags


liggitt:
As part of sync-tags, we update go.mod files to refer to tagged releases of dependencies. This results in a tag-specific commit/SHA for most repos. This is desired, so that go resolves that sha to that specific semver tag.

If a repo has no peer dependencies (like k8s.io/apimachinery), no change is required to the go.mod file, so it is possible for multiple tags to be created pointing to the same commit/sha (like [v0.17.0](https://github.com/kubernetes/apimachinery/releases/tag/v0.17.0) and [v0.17.1-beta.0](https://github.com/kubernetes/apimachinery/releases/tag/v0.17.1-beta.0)).

While this is functionally correct (nothing breaks), it is confusing if go resolves a sha or non-semver tag (like kubernetes-1.17.0) to an unexpected semver tag (like v0.17.1-beta.0), even if it is equivalent content.

In sync-tags, when syncing a semver tag, if we also added a comment like "# Corresponds to Kubernetes v1.17.0", this would force a diff in all go.mod files, force a tag-specific commit, and reduce confusion in sha/non-semver resolution.

xref https://github.com/kubernetes/kubernetes/issues/84372#issuecomment-564562164

nikhita:
/remove-lifecycle rotten

this would be nice to have
/assign

benmoss:
I just ran into this while trying to automate vendoring via the script in https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597

Here's a repro of the unexpected behavior:

```
$ go mod download -json k8s.io/controller-manager@kubernetes-1.19.1
{
        "Path": "k8s.io/controller-manager",
        "Version": "v0.19.3-rc.0",
        "Info": "/home/cesium/go/pkg/mod/cache/download/k8s.io/controller-manager/@v/v0.19.3-rc.0.info",
        "GoMod": "/home/cesium/go/pkg/mod/cache/download/k8s.io/controller-manager/@v/v0.19.3-rc.0.mod",
        "Zip": "/home/cesium/go/pkg/mod/cache/download/k8s.io/controller-manager/@v/v0.19.3-rc.0.zip",
        "Dir": "/home/cesium/go/pkg/mod/k8s.io/controller-manager@v0.19.3-rc.0",
        "Sum": "h1:j4Np5Ap5bUm7dWf+YpdtEO7onPh+bapRVyvDuX7+moA=",
[...]

Issue #187, opened 2019-06-20
sync-tags adds an empty commit when pseudoversion doesn't change


nikhita:
Example: https://github.com/kubernetes/api/commit/7cf5895f2711098d7d9527db0a4a49fb0dff7de2

The pseudoversion for apimachinery doesn't change in the above case, but sync-tags still adds an empty commit. This isn't harmful, but is redundant. 

We should be able to avoid add that by checking the version by `go list -m -json k8s.io/apimachinery` and the expected pseudoversion.

/assign
will look over the weekend, don't have time to work on this + test it before then

liggitt:
given https://github.com/kubernetes/publishing-bot/issues/215, I'm not sure this is still true, and if it is, I think we may actually want this behavior

PR #227, opened 2020-06-08
sync-tags: run go mod tidy after updating deps in go.mod

nikhita:
Consider a repo (like component-base) that depends on apimachinery, api
(which depnds on apimachinery) and client-go (which depends on
apimachinery and api). In the current case, sync-tags updates the tags
in component-base's go.mod in the following way:

1. Update tag for apimachinery, run `go mod tidy`
2. Update tag for api, run `go mod tidy`
3. Update tag for client-go, run `go mod tidy`

If we remove a package from api and it's references in client-go in k/k,
step 2 fails while running `go mod tidy` because the go.mod file still
contains the old client-go version i.e. client-go still references
the removed package.

To fix this, we should update tags for all dependencies and _then_ run
`go mod tidy`. We also use this approach in util.sh while syncing branches
using pseudoversions so we don't face any error there.

https://github.com/kubernetes/publishing-bot/blob/ff2d8bb9188faeb5b4dd9369cae300e6c654c015/artifacts/scripts/util.sh#L857-L868

---

This fixes errors like (due to https://github.com/kubernetes/kubernetes/commit/e857adbdfdba795ceca870f194d8d8a296bbdc21):

```
k8s.io/component-base/metrics/prometheus/clientgo/leaderelection imports
    k8s.io/client-go/tools/leaderelection tested by
    k8s.io/client-go/tools/leaderelection.test imports
    k8s.io/client-go/kubernetes/fake imports
    k8s.io/api/auditregistration/v1alpha1: module k8s.io/api@latest found (v0.19.0-beta.1, replaced by k8s.io/api@v0.19.0-beta.1), but does not contain package k8s.io/api/auditregistration/v1alpha1
```

### DIFF
--- a/cmd/sync-tags/gomod.go
+++ b/cmd/sync-tags/gomod.go
@@ -97,22 +97,6 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag 
 			return changed, fmt.Errorf("unable to pin %s in the replace section of go.mod to %s: %v", depPkg, pseudoVersionOrTag, err)
 		}
 
-		downloadCommand2 := exec.Command("go", "mod", "download")
-		downloadCommand2.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPRIVATE=%s", depPackages), "GOPROXY=https://proxy.golang.org")
-		downloadCommand2.Stdout = os.Stdout
-		downloadCommand2.Stderr = os.Stderr
-		if err := downloadCommand2.Run(); err != nil {
-			return changed, fmt.Errorf("error running go mod download for pseudo-version %s for %s: %v", pseudoVersionOrTag, depPkg, err)
-		}
-
-		tidyCommand := exec.Command("go", "mod", "tidy")
-		tidyCommand.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPROXY=file://%s/pkg/mod/cache/download", os.Getenv("GOPATH")))
-		tidyCommand.Stdout = os.Stdout
-		tidyCommand.Stderr = os.Stderr
-		if err := tidyCommand.Run(); err != nil {
-			return changed, fmt.Errorf("unable to run go mod tidy for %s at %s: %v", depPkg, rev, err)
-		}
-
 		found[dep] = true
 		fmt.Printf("Bumping %s in go.mod to %s.\n", depPkg, rev)
 		changed = true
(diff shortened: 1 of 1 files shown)


nikhita:
/hold
IMO we should do a dry run or a nightly test before merging this

cmd/sync-tags/gomod.go
```diff
@@ -123,6 +107,24 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag
...
+
+	downloadCommand2 := exec.Command("go", "mod", "download")
+	downloadCommand2.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPRIVATE=%s", depPackages), "GOPROXY=https://proxy.golang.org")
+	downloadCommand2.Stdout = os.Stdout
+	downloadCommand2.Stderr = os.Stderr
+	if err := downloadCommand2.Run(); err != nil {
+		return changed, fmt.Errorf("error running go mod download: %v", err)
+	}
+
+	tidyCommand := exec.Command("go", "mod", "tidy")
```
nikhita, line 119: IIRC we decided to run `go mod tidy` after each update because we wanted to ensure that tags were correctly packaged and didn't contain errors.

If we run `go mod tidy` after all dependency updates, go will still check that all new tags are correctly packaged but I think we can't be sure if it removed any lines (not really sure where we might hit this case though)...

Having said that, we can't be sure even in the current case because subsequent `go mod tidy`s might still remove our updates, right? :thinking: 

Slightly tangential, but we might also want to remove `changed` and consider it as true always to help with https://github.com/kubernetes/publishing-bot/issues/187#issuecomment-564595877 and https://github.com/kubernetes/publishing-bot/issues/215. Thoughts?

https://github.com/kubernetes/publishing-bot/blob/ff2d8bb9188faeb5b4dd9369cae300e6c654c015/cmd/sync-tags/main.go#L330-L336

sttts:
> IMO we should do a dry run or a nightly test before merging this

This, please.

Am 80% confident the change is ok.

liggitt:
agree the change looks good, and agree we want a test run

cmd/sync-tags/gomod.go
```diff
@@ -123,6 +107,24 @@ func updateGomodWithTaggedDependencies(tag string, depsRepo []string, semverTag
...
+	tidyCommand.Env = append(os.Environ(), "GO111MODULE=on", fmt.Sprintf("GOPROXY=file://%s/pkg/mod/cache/download", os.Getenv("GOPATH")))
+	tidyCommand.Stdout = os.Stdout
+	tidyCommand.Stderr = os.Stderr
+	if err := tidyCommand.Run(); err != nil {
+		return changed, fmt.Errorf("unable to run go mod tidy: %v", err)
+	}
+	fmt.Printf("Completed running go mod tidy for %s.\n", tag)
```
nikhita, line 126: Update to add `\n` here

dims:
/approve
/lgtm

please feel free to remove /hold

nikhita:
Pushed latest image to https://console.cloud.google.com/gcr/images/k8s-staging-publishing-bot/GLOBAL/k8s-publishing-bot?gcrImageListsize=30

PR #228, opened 2020-06-15
sync-tags: use go-git for fetching tags

nikhita:
If git is used directly while fetching tags, go-git doesn't recognize
annotated tags consistently. For example, there are instances where
it considers `refs/tags/upstream/*` as lightweight tags even though
`git for-each-ref refs/tags` clearly shows them as annotated tags.

This commits updates the fetch operation to use go-git directly to avoid
any inconsistencies while fetching and storing tags. Using go-git
removes this inconsistency and all annotated tags are recognized
properly.

Also, go-git now supports `*` in refspecs - https://github.com/src-d/go-git/blob/v4.9.1/config/refspec.go#L11.

---

We came across this error while deploying the bot recently. The bot doesn't recognize `refs/tags/upstream/*` as annotated, which makes sync-tags think that there are no new upstream tags and it skips syncing tags entirely. See [slack thread](https://kubernetes.slack.com/archives/C2C40FMNF/p1591792990209100?thread_ts=1591701817.194100&cid=C2C40FMNF) for more context.

/assign @dims @sttts

### DIFF
--- a/cmd/sync-tags/main.go
+++ b/cmd/sync-tags/main.go
@@ -32,8 +32,10 @@ import (
 	"github.com/golang/glog"
 	"github.com/renstrom/dedent"
 	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/sideband"
 
 	"k8s.io/publishing-bot/pkg/cache"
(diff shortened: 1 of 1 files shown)


nikhita:
/hold
for testing

nikhita:
This is an example output for `git show-ref --tags` for apimachinery after creating new tags `v0.19.0-beta.1` and `kubernetes-1.19.0-beta.1` - https://gist.github.com/nikhita/d299aeda503edd72df58d6630967bea4

sttts:
Sgtm.

nikhita:
Nightly test run looks good: https://github.com/kubernetes-nightly/component-base/blob/v0.19.0-beta.2/go.mod :tada: 

/hold cancel

PR #231, opened 2020-07-13
util.sh: avoid side-effect from go mod download in dependencies

sttts:


nikhita:
hit lgtm and remembered that we also do a `go mod download` in a similar case in sync-tags/gomod.go

PR #232, opened 2020-07-16
Move default go version to rules file

nikhita:
Today, the `defaultGoVersion` is hardcoded in the publishing-bot. This
means that whenever k/k undergoes a go version bump, we need to manually
update the hardcoded version in the bot.

To ensure that we can bump the go version for publishing-bot along with
other places in k/k, this commit moves `defaultGoVersion` to the rules
file.

This commit introduces the following changes:

- A new optional field `default-go-version` is introduced in the rules
file.

- To avoid breaking existing users, if the `default-go-version` field is
not specified in the rules file, the bot defaults to
`deprecatedDefaultGoVersion` (go 1.14.4). Since go stops supporting
Go 1.x once Go 1.x+2 is released, the defaulting to
`deprecatedDefaultGoVersion` will be removed once go 1.16 is released
and users will be required to specify the `default-go-version` field.

- We now validate that all go versions specified are valid semver
versions.

- Installation of the default go version is moved from `init-repo`
to the `publishing-bot` binary. This is needed because we need to update
the source repo and read the rules file in [`Run()`](https://github.com/kubernetes/publishing-bot/blob/4437c32dbde906dd4a9d7ab2ca0cea6b72dc6146/cmd/publishing-bot/publisher.go#L393-L421) (in publisher.go)
before installing the defaultGoVersion.

- `init-repo` no longer installs godep and dep. We moved away from
godep and dep a long time ago. We use [`godeps-gen`](https://github.com/kubernetes/publishing-bot/blob/master/cmd/godeps-gen/main.go) to generate
the fake Godeps.json file but the Godeps.json file in this case is
generated by manipulating the file directly, and not by using the `godep`
binary.

### DIFF
--- a/cmd/init-repo/main.go
+++ b/cmd/init-repo/main.go
@@ -29,12 +29,6 @@ import (
 	yaml "gopkg.in/yaml.v2"
 
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
-	"k8s.io/publishing-bot/pkg/golang"
-)
-
-const (
-	depCommit   = "7c44971bbb9f0ed87db40b601f2d9fe4dffb750d"
-	godepCommit = "tags/v80"
 )
 
 var (
@@ -60,8 +54,6 @@ func main() {
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	rulesFile := flag.String("rules-file", "", "the file with repository rules")
 	targetOrg := flag.String("target-org", "", `the target organization to publish into (e.g. "k8s-publishing-bot")`)
-	skipGodep := flag.Bool("skip-godep", false, `skip godeps installation and godeps-restore`)
-	skipDep := flag.Bool("skip-dep", false, `skip 'dep'' installation`)
 
 	flag.Usage = Usage
 	flag.Parse()
@@ -139,18 +131,7 @@ func main() {
 		glog.Fatalf("Failed to create source repo directory %s: %v", BaseRepoPath, err)
 	}
 
-	if err := golang.InstallDefaultGoVersion(); err != nil {
-		glog.Fatalf("Failed to install default go version: %v", err)
-	}
-
-	if !*skipGodep {
-		installGodeps()
-	}
-	if !*skipDep {
-		installDep()
-	}
-
-	cloneSourceRepo(cfg, *skipGodep)
+	cloneSourceRepo(cfg)
 	for _, rule := range rules.Rules {
 		cloneForkRepo(cfg, rule.DestinationRepository)
 	}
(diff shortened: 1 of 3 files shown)


nikhita:
Tested both with and without the `default-go-version` field in rules file. :+1: 

/assign @sttts @dims

cmd/init-repo/main.go
```diff
@@ -60,8 +54,6 @@ func main() {
 	repoOrg := flag.String("source-org", "", "the name of the source repository organization, (eg. kubernetes)")
 	rulesFile := flag.String("rules-file", "", "the file with repository rules")
 	targetOrg := flag.String("target-org", "", `the target organization to publish into (e.g. "k8s-publishing-bot")`)
-	skipGodep := flag.Bool("skip-godep", false, `skip godeps installation and godeps-restore`)
```
nikhita, line 63: Just as a note...

This flag is only used for `init-repo`. For specifying if godeps should be generated or not, `skip-godeps` is used in the rules file:

https://github.com/kubernetes/publishing-bot/blob/4437c32dbde906dd4a9d7ab2ca0cea6b72dc6146/cmd/publishing-bot/config/rules.go#L69

https://github.com/kubernetes/publishing-bot/blob/b1513c6d5c7dbbec57fbc1c3f3e339c24e16aba7/cmd/publishing-bot/publisher.go#L291-L293

This PR only removes the `skip-godeps` flag from `init-repo`. It does *not* remove the `skip-godeps` option from rules.

cmd/init-repo/main.go
```diff
@@ -139,18 +131,7 @@ func main() {
 		glog.Fatalf("Failed to create source repo directory %s: %v", BaseRepoPath, err)
 	}
 
-	if err := golang.InstallDefaultGoVersion(); err != nil {
```
sttts, line 142: this is gone?

cmd/publishing-bot/config/rules.go
```diff
@@ -46,7 +47,7 @@ func (c Source) String() string {
 
 type BranchRule struct {
 	Name string `yaml:"name"`
-	// a (full) version string like 1.10.2.
+	// a (full) semver version string like 1.10.2.
```
sttts, line 50: so 1.10 works too?

cmd/init-repo/main.go
nikhita: This is moved to `InstallGoVersions`: https://github.com/kubernetes/publishing-bot/pull/232/files#diff-ca3cd93ee51f0f2890610d4399786b2bR44-R48

InstallGoVersions() is called in construct():

https://github.com/kubernetes/publishing-bot/blob/b1513c6d5c7dbbec57fbc1c3f3e339c24e16aba7/cmd/publishing-bot/publisher.go#L196-L201

cmd/publishing-bot/config/rules.go
nikhita: Updated so that 1.10 works but 1.10.0 won't. PTAL.

cmd/init-repo/main.go
sttts: Ack.

cmd/publishing-bot/config/rules.go
nikhita: I don't really like this hack though...

sttts: Actually there is no non-full semver version string. It's pretty clearly defined in https://semver.org/.

sttts: so which do we want? Semver or not semver? :-)

nikhita: I updated it to say "valid Go version" :)

https://github.com/kubernetes/publishing-bot/pull/232/files#diff-d058080bd6d5b27f3f6ff87d286765e8R50

cmd/publishing-bot/config/rules.go
```diff
@@ -142,10 +147,49 @@ func validateRepoOrder(rules *RepositoryRules) (errs []error) {
...
+		return fmt.Errorf("go version %s should not contain the .0 suffix", version)
+	}
+
+	// the semver library requires a patch version, so append a .0
+	// to be able to validate the major/minor versions
+	if len(parts) == 2 {
+		s = s + ".0"
+	}
+	if _, err := semver.Parse(s); err != nil {
+		return fmt.Errorf("go version %s must be a valid semver version: %v", version, err)
```
nikhita, line 183: oops will update here

nikhita: Done

nikhita:
@sttts This is ready for review :)

nikhita:
Deployed

PR #235, opened 2020-09-01
Fix config value does not work

daixiang0:
Since flag has default value(never empty), so config value never work.

Signed-off-by: Xiang Dai <long0dai@foxmail.com>


cmd/publishing-bot/main.go
```diff
@@ -94,9 +96,12 @@ func main() {
 	if *rulesFile != "" {
 		cfg.RulesFile = *rulesFile
 	}
-	if *basePublishScriptPath != "" {
+	if *basePublishScriptPath != defaultPublishScriptPath {
 		cfg.BasePublishScriptPath = *basePublishScriptPath
+	} else if cfg.BasePublishScriptPath == "" {
+		cfg.BasePublishScriptPath = defaultPublishScriptPath
```
nikhita, line 102: Since command line flags are meant to override config values, we can simplify this:

```go
	if cfg.BasePublishScriptPath == "" {
		cfg.BasePublishScriptPath = *basePublishScriptPath
	}
```

nikhita:
While we are here, can you also update the comment on the config to `./publish_scripts` instead of `./publishing_scripts`?

https://github.com/kubernetes/publishing-bot/blob/85ef43ebc715b017324cbd030fbde06f31a2f681/cmd/publishing-bot/config/config.go#L52

cmd/publishing-bot/main.go
daixiang0: It does not cover the case that custom config value + custom flag value.

daixiang0:
@nikhita done.

daixiang0:
ping @nikhita

PR #236, opened 2020-09-01
Update rule example

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

### DIFF
--- a/configs/example-rules-configmap.yaml
+++ b/configs/example-rules-configmap.yaml
@@ -12,8 +12,6 @@ data:
     # - BUILD
     # - "*/BUILD"
 
-    #             RecursiveDeletePatterns []string `yaml:"recursive-delete-patterns"`
-
     # Skip update/fix godeps
     # skip-godeps: true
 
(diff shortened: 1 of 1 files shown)


configs/example-rules-configmap.yaml
```diff
@@ -11,6 +11,23 @@ data:
     recursive-delete-patterns:
     # - BUILD
     # - "*/BUILD"
+
+    #             RecursiveDeletePatterns []string `yaml:"recursive-delete-patterns"`
```
nikhita, line 15: This looks like a go struct field so we should remove this from the yaml config

configs/example-rules-configmap.yaml
```diff
@@ -11,6 +11,23 @@ data:
...
+
+    # Skip update/fix gomod
+    # skip-gomod: true
+
+    # Skip sync tags
+    # skip-tags: true
+
+    # a valid go version string like 1.10.2 or 1.10
+    # if GoVersion is not specified in RepositoryRule,
+    # DefaultGoVersion is used.
```
nikhita, line 28:
```suggestion
    # if the go version is not specified in rules,
    # default-go-version is used.
```

PR #238, opened 2020-10-14
Skip checkout of deps if their look-up table is empty

nikhita:
If a dependency's look-up table is empty, we will get an error while
trying to read the `kube-commits-${dep}-${branch}` file.

To avoid getting this error, this commit adds a check for the file size
before reading the file.

---

Note to self -  a dependency's look-up table can be empty if it does
not have merge commits in the filtered branch.

This case is possible because:

- `git filter branch` drops fast-forward merges which don't contain
diffs.
- While constructing the tree to create the dropped merge commit, it is
possible that the second parent of the merge commit is also `HEAD`
of the filtered branch. Since both parents would be then be `HEAD`,
we end up with a tree with no merge commits as first parents.


nikhita:
/assign @sttts 

Do you think this needs a smoke test?

nikhita:
Thanks @dims!

Going to deploy this tomorrow morning my time so that I'm awake when it's running :)

PR #244, opened 2021-01-26
gomod-zip: ignore submodules in zip archive

nikhita:
Currently while zipping modules using the `gomod-zip` tool, if a repo
contains submodules, we include them in the zip archive.

However, when we do a `go mod download`, go skips submodules from the
zip archive. This means that the zip file created using `gomod-zip`
is no longer accurate.

This also causes the following error when go tries to read the zip file
(during `go mod tidy`, etc) and notices submodules in the zip archive:

```
-> unzip /go-workspace/pkg/mod/cache/download/k8s.io/code-generator/@v/v0.0.0-20210126111002-a3ae865d6348.zip: found go.mod file not in module root directory (k8s.io/code-generator@v0.0.0-20210126111002-a3ae865d6348/examples/go.mod)
unzip /go-workspace/pkg/mod/cache/download/k8s.io/code-generator/@v/v0.0.0-20210126111002-a3ae865d6348.zip: found go.mod file not in module root directory (k8s.io/code-generator@v0.0.0-20210126111002-a3ae865d6348/examples/go.mod)
```

This specific error occurs because submodules were recently added to
code-generator in https://github.com/kubernetes/kubernetes/pull/98197.

To fix this issue, this commit ignores submodules while creating the
zip archive. The code is derived from https://github.com/golang/go/blob/b373d31c25e58d0b69cff3521b915f0c06fa6ac8/src/cmd/go/internal/modfetch/coderepo.go#L572.
Note that since we already know the package name, this commit removes
some complexity for deriving the package name from the original go code.

---

Verified this using an example repo - https://gist.github.com/nikhita/f2fc3096e8900133de2645370fdab9c6

/hold
/cc @dims @sttts @liggitt 
/assign @sttts

### DIFF
--- a/cmd/gomod-zip/zip.go
+++ b/cmd/gomod-zip/zip.go
@@ -133,6 +133,30 @@ func main() {
 	defer dst.Close()
 	zw := zip.NewWriter(dst)
 
+	isSubmodule := map[string]bool{}
+	for _, zf := range zr.File {
+		if zf.Name == prefix+"go.mod" {
+			continue
+		}
+		dir, file := path.Split(zf.Name)
+		if file == "go.mod" {
+			isSubmodule[dir] = true
+		}
+	}
+
+	inSubmodule := func(name string) bool {
+		for {
+			dir, _ := path.Split(name)
+			if len(dir) == 0 {
+				return false
+			}
+			if isSubmodule[dir] {
+				return true
+			}
+			name = strings.TrimSuffix(dir, "/")
+		}
+	}
+
 	for _, zf := range zr.File {
 		// Skip symlinks (golang.org/issue/27093)
 		if !zf.FileInfo().Mode().IsRegular() {
@@ -154,6 +178,10 @@ func main() {
 		}
 		// don't consider vendored packages
 		if isVendoredPackage(name) {
+			continue
+		}
+		// don't consider submodules
+		if inSubmodule(zf.Name) {
 			continue
 		}
 		// make sure we have lower-case go.mod


cmd/gomod-zip/zip.go
```diff
@@ -133,6 +133,27 @@ func main() {
...
+		dir, file := path.Split(zf.Name)
+		if file == "go.mod" {
+			haveGoMod[dir] = true
+		}
+	}
+
+	inSubmodule := func(name string) bool {
+		for {
+			dir, _ := path.Split(name)
+			if len(dir) <= len(prefix) {
```
dims, line 147: dir is a array of strings and prefix is a string ... no? (trying to understand)

nikhita: `dir` is also a string - https://golang.org/pkg/path/#Split

This condition is to ensure we ignore the root go.mod file when searching for submodules.

sttts:
SGTM. Leaving review to @dims

nikhita:
For visibility, this PR does _not_ add multi-module functionally to the publishing-bot. The bot will still publish a wrong go.mod file for `examples/go.mod` - https://github.com/kubernetes/kubernetes/pull/98197#issuecomment-767561731. The root go.mod will be published correctly.

This PR just unblocks publishing of the root go.mod file. We _probably_ don't care about the published version of `examples/go.mod` so just unblocking publishing of root go.mod should be fine. However, if we decide to support multimodules, it will require additional complexity to be added to the bot.

cmd/gomod-zip/zip.go
```diff
@@ -133,6 +133,27 @@ func main() {
 	defer dst.Close()
 	zw := zip.NewWriter(dst)
 
+	haveGoMod := make(map[string]bool)
+	for _, zf := range zr.File {
+		dir, file := path.Split(zf.Name)
+		if file == "go.mod" {
```
liggitt, line 139: would it be simpler to only track submodules?

```
isSubmodule := map[string]bool{}
for _, zf := range zr.File {
  if zf.Name == prefix + "go.mod" {
    continue
  }
  dir, file := path.Split(zf.Name)
  if file == "go.mod" {
    isSubmodule[dir] = true
  }
}
```

cmd/gomod-zip/zip.go
liggitt: if we only track submodules (see above comment), this could be simplified to `if len(dir) == 0`

cmd/gomod-zip/zip.go
```diff
@@ -133,6 +133,30 @@ func main() {
...
+	inSubmodule := func(name string) bool {
+		for {
+			dir, _ := path.Split(name)
+			if len(dir) <= len(prefix) {
+				return false
+			}
+			if isSubmodule[dir] {
+				return true
+			}
+			name = dir[:len(dir)-1]
```
liggitt, line 156: `name = strings.TrimSuffix(dir, "/")` might be simpler to read

cmd/gomod-zip/zip.go
nikhita: Updated :+1:

cmd/gomod-zip/zip.go
nikhita: Done

cmd/gomod-zip/zip.go
nikhita: Done

nikhita:
@dims please cancel the hold if this lgty :)

cmd/gomod-zip/zip.go
dims: whoops. gotcha!

dims:
/hold cancel

thanks @nikhita

PR #245, opened 2021-02-04
rules: fix go version validation

nikhita:
Ref: https://github.com/kubernetes/kubernetes/pull/98572#discussion_r567076244

The current validation fails for the rc tag because go versions
don't follow semver. For example, 1.16rc1 is a valid go version
but the publishing-bot validation fails.

To fix this issue, this commit validates using a regular expression.
Regex is derived from https://github.com/kubernetes-sigs/kubebuilder/blob/3c8e370d49c58179037d69f926318093f82f68d2/pkg/plugins/internal/util/go_version.go#L57.

/assign @dims 
cc @sttts @justaugustus

### DIFF
--- a/cmd/publishing-bot/config/rules.go
+++ b/cmd/publishing-bot/config/rules.go
@@ -7,10 +7,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


dims:
/approve
/lgtm

thanks @nikhita

PR #254, opened 2021-04-21
gomod-zip: use golang.org/x/mod/zip for creating zip file

nikhita:
The zip file created by gomod-zip no longer matches the zip file created
by go mod directly. The initial gomod-zip code copied Go's code
(at https://github.com/golang/go/blob/4d56576ec02e3e3b4459bc23eb1601e6e96f3cfc/src/cmd/go/internal/modfetch/coderepo.go#L940)
but has diverged now.

Go now extracts the logic to create the zip file in `golang.org/x/mod/zip`:
https://github.com/golang/mod/blob/master/zip/zip.go so this commit
updates gomod-zip to use modzip.

The version of `golang.org/x/mod` is pinned to the version that
go uses - https://github.com/golang/go/blob/952187af12485eb665ae122f6d0bdb36e4a11ed7/src/cmd/go.mod#L9
i.e. `v0.4.3-0.20210409134425-858fdbee9c24`.
We will need to update this version whenever go changes it to ensure that
we don't create different zip files. Otherwise, this will lead to a
checksum error.

/assign @sttts @dims


nikhita:
/hold
I'd like to test this against all our branches/go versions to ensure this doesn't have different behavior across different go versions

nikhita:
> We will need to update this version whenever go changes it to ensure that
we don't create different zip files. Otherwise, this will lead to a
checksum error.

We can get rid of this when (and if) go implements https://github.com/golang/go/issues/33312

dims:
the unified diff looks awful :) the side-by-side diff shows the new code looks much better. thanks @nikhita 

/approve
/lgtm

nikhita:
This fixed the error for `release-1.19` and `release-1.20` branches, but the bot still tripped for `release-1.21` because the current published go.sum already contains wrong hash values. Created https://github.com/kubernetes/publishing-bot/pull/255 to fix this

PR #255, opened 2021-04-26
util.sh: regenerate go.sum to fix incorrect hash values

nikhita:
[gomod-zip](https://github.com/kubernetes/publishing-bot/tree/master/cmd/gomod-zip) copied go's zip creation code but they had diverged.
due to this, gomod-zip created different zip files in the cache
compared to what go mod download would create (fixed in https://github.com/kubernetes/publishing-bot/pull/254)

From Go 1.15.11 and Go 1.16.3, go automatically derives the ziphash
from the zip file in the cache - https://github.com/golang/go/issues/44812.
This meant that go added incorrect hash values to go.sum because these
were derived from the zip files produced by the diverged gomod-zip code.

So remove go.sum here and regenerate again using go mod download and go mod tidy
to include the right hash values.

Note - this line should be removed after one clean publishing-bot run to be able to catch checksum
errors from `go mod download` in the future.

---

Fixes broken publishing-bot. We get this error for component-base `release-1.21` now:

```
    	+ GO111MODULE=on
    	+ GOPRIVATE=k8s.io/apimachinery,k8s.io/api,k8s.io/client-go
    	+ GOPROXY=https://proxy.golang.org
    	+ go mod download
    	verifying k8s.io/client-go@v0.0.0-20210329112528-6be0785e0e32: checksum mismatch
    		downloaded: h1:Qu7DBMSFCtY/i0Q5reJbTOgwbPZqOQSdLszDn05zbBA=
    		go.sum:     h1:rypE67YonSf09C58AflueN1+kwxCCxJb+f6hkxjSECI=

    	SECURITY ERROR
    	This download does NOT match an earlier download recorded in go.sum.
    	The bits may have been replaced on the origin server, or an attacker may
    	have intercepted the download attempt.

    	For more information, see 'go help module-auth'.
```

Note - `go mod tidy` on the published component-base repo's master also produces a checksum error today.

/hold
/cc @sttts @dims @liggitt

### DIFF
--- a/artifacts/scripts/util.sh
+++ b/artifacts/scripts/util.sh
@@ -859,6 +859,21 @@ update-deps-in-gomod() {
 
     GO111MODULE=on go mod edit -json | jq -r '.Replace[]? | select(.New.Path | startswith("../")) | "-dropreplace \(.Old.Path)"' | GO111MODULE=on xargs -L 100 go mod edit -fmt
     
+    # TODO(nikhita): remove this after go.sum values are fixed.
+    #
+    # gomod-zip copied go's zip creation code but they had diverged.
+    # due to this, gomod-zip created different zip files in the cache
+    # compared to what go mod download would create.
+    #
+    # From Go 1.15.11 and Go 1.16.3, go automatically derives the ziphash
+    # from the zip file in the cache - https://github.com/golang/go/issues/44812.
+    # This meant that go added incorrect hash values to go.sum because these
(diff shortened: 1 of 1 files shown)


nikhita:
Verified for component-base `release-1.21` - a new commit is published which fixes the hash in go.sum - https://github.com/prowtest/component-base/commit/014c101d6de79a70164460f5fa3a7e31af76c0ce.

Also verified that `go mod tidy` is green after this new commit is published.

PR #261, opened 2021-06-02
Stop publishing synthetic Godeps.json files

nikhita:
Ref: https://github.com/kubernetes/kubernetes/pull/102454#issuecomment-851912151

From https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/917-go-modules/README.md#publish-staging-component-modules-to-individual-repositories:
> Generate synthetic Godeps.json files containing the SHA or git tag of
> module dependencies, for consumption by downstream consumers using
> dependency management tools like glide. Continue publishing these at
> least until our minimum supported version of go defaults to enabling
> module support (currently targeted for go 1.13, which is approximately
> Kubernetes 1.16-1.17 timeframe).

Now that Go 1.16 is in module-aware mode by default, it should be safe
to remove support for publishing synthetic `Godeps.json` files.


nikhita:
/hold

@sttts IIRC openshift was one of the primary consumers of this file. Can you confirm that this is safe to remove?

Alternatively, we could just use `skip-godeps: true` for k/k and keep the code for publishing synthetic `Godeps.json` files in the publishing-bot.

nikhita:
/hold cancel

based on https://kubernetes.slack.com/archives/CHGFYJVAN/p1623180805091300?thread_ts=1623163166.090300&cid=CHGFYJVAN

PR #278, opened 2021-12-07
Script to fork/create repositories needed by publishing-bot

dims:
often we add new repos in staging and it's a manual process to create the repositories. let's automate that.

```hack/create-repos.sh kubernetes-nightly```

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/hack/create-repos.sh
+++ b/hack/create-repos.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")
+source "${SCRIPT_ROOT}/repos.sh"
+
+if [ "$#" = 0 ] || [ "$#" -gt 2 ]; then
+    echo "usage: $0 [source-github-user-name] dest-github-user-name"
+    echo
+    echo "This connects to git@github.com:<from>/<repo>. Set GITHUB_HOST to access git@<GITHUB_HOST>:<from>/<repo> instead."
+    exit 1
+fi
+
+FROM="kubernetes"
+TO="${1}"
+if [ "$#" -ge 2 ]; then
+    FROM="${TO}"
+    TO="${2}"
+fi
+GITHUB_HOST=${GITHUB_HOST:-github.com}
+
+repo_count=${#repos[@]}
+
+# safety check
+if [ "${TO}" = "kubernetes" ]; then
+    echo "Cannot operate on kubernetes directly" 1>&2
+    exit 1
+fi
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: hack/repos.sh [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+    kubectl
+    kubelet
+    legacy-cloud-providers
+    metrics
+    mount-utils
+    pod-security-admission
+    sample-apiserver
+    sample-cli-plugin
+    sample-controller
+)
```
justaugustus, line 50:
```suggestion
)

```

hack/repos.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
justaugustus, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/create-repos.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
justaugustus, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/create-repos.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# Copyright 2017 The Kubernetes Authors.
```
justaugustus, line 3:
```suggestion
# Copyright 2021 The Kubernetes Authors.
```

hack/repos.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Copyright 2017 The Kubernetes Authors.
```
justaugustus, line 3:
```suggestion
# Copyright 2021 The Kubernetes Authors.
```

hack/create-repos.sh
```diff
@@ -0,0 +1,69 @@
...
+if [ "${TO}" = "kubernetes" ]; then
+    echo "Cannot operate on kubernetes directly" 1>&2
+    exit 1
+fi
+
+destination_repos=( $(curl -ks https://api.github.com/orgs/${TO}/repos | jq ".[].name" | tr -d '"') )
+destination_repo_count=${#destination_repos[@]}
+
+# Checks if you are logged in. Will error/bail if you are not.
+gh auth status
```
justaugustus, line 51: We should make mention somewhere that `gh` is a requirement in the environment and link to the docs: https://cli.github.com/

hack/create-repos.sh
```diff
@@ -0,0 +1,69 @@
...
+if [ "$#" -ge 2 ]; then
+    FROM="${TO}"
+    TO="${2}"
+fi
+GITHUB_HOST=${GITHUB_HOST:-github.com}
+
+repo_count=${#repos[@]}
+
+# safety check
+if [ "${TO}" = "kubernetes" ]; then
```
justaugustus, line 42: Maybe an env var for forbidden destinations?

justaugustus (changes requested):
@dims -- Left a few notes!

hack/create-repos.sh
dims: this snippet is from the other shell script, let's update both at the same time after this PR lands please.

hack/create-repos.sh
dims: Done!

justaugustus (approved):
Thanks @dims!

hack/repos.sh
```diff
@@ -0,0 +1,50 @@
...
+    kube-scheduler
+    kubectl
+    kubelet
+    legacy-cloud-providers
+    metrics
+    mount-utils
+    pod-security-admission
+    sample-apiserver
+    sample-cli-plugin
+    sample-controller
```
sttts, line 49: why don't we pull these from k/k?

dims: @sttts we could ... but this turns out to be a super set of all the repos that exist in all the current stable branches (gives that we may be introducing new repos and we may be moth-balling others). So a bit more complicated than `ls` in the correct directory under k/k master.

PR #285, opened 2022-01-17
Add config option to remove .gitattributes files

nikhita:
If a staging repo has .gitattributes files containing the `export-subst`
attribute ([example](https://github.com/kubernetes/kubernetes/blob/b6c06a95d7ff262a876b95a1035eaf478a7cb961/staging/src/k8s.io/client-go/pkg/version/.gitattributes)), then git expands the specified placeholders
when git archive is used.

When a published repo is downloaded from GitHub, GitHub does a
"git archive" under the hood. This means that the placeholders get
replaced by their relevant values. This type of "git archive"
application sometimes leads to undesired values. See the example below.

- In client-go, [line 59 in `pkg/version/base.go`](https://github.com/kubernetes/kubernetes/blob/b6c06a95d7ff262a876b95a1035eaf478a7cb961/staging/src/k8s.io/client-go/pkg/version/base.go#L59)
is expanded on a git archive. This line is needed as a fallback to
inject k8s version info for client-go when this info is not provided
via ldflags during builds.

- However, when k/client-go is vendored, the line gets expanded to _the
commit of the project vendoring client-go_ -- which is not helpful at
all! This also means that the vendored client-go will now contain
different (expanded) commit shas for different projects.

- To ensure reproducibility of source, this commit adds an option to
remove the `.gitattributes` files in all published repos.
Additionally, when client-go is used as a library, we don't care about
the line being expanded to inject version info so it is also safe to
remove these files.

The commit adds this feature as a config option in `rules.yaml` so we
can control publishing of `.gitattributes` files for each repo.

---

Ref:
- https://github.com/containerd/containerd/issues/6387#issuecomment-1010019352 (issue found in containerd, cri-o and podman)
- https://github.com/kubernetes/kubernetes/issues/99376#issuecomment-1009881488

---

I have tested this on master for client-go and component-base. The result will look something like this for:
1. client-go - https://github.com/prowtest/client-go/commits/master
2. component-base - https://github.com/prowtest/component-base/commits/master

/hold
Adding a hold because I'd like to ensure tags work fine too. Will do that later this week but wanted to get the PR out for review.

/assign @sttts @dims

### DIFF
--- a/artifacts/scripts/construct.sh
+++ b/artifacts/scripts/construct.sh
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 14 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_packages kubernetes_remote subdirectory source_repo_org source_repo_name base_package is_library recursive_delete_pattern skip_tags last_published_upstream_hash"
+if [ ! $# -eq 15 ]; then
+    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_packages kubernetes_remote subdirectory source_repo_org source_repo_name base_package is_library recursive_delete_pattern skip_tags last_published_upstream_hash remove_git_attributes"
     exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


dims:
@kubernetes/release-engineering please review!

artifacts/scripts/construct.sh
```diff
@@ -37,8 +37,8 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [ ! $# -eq 14 ]; then
-    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_packages kubernetes_remote subdirectory source_repo_org source_repo_name base_package is_library recursive_delete_pattern skip_tags last_published_upstream_hash"
+if [ ! $# -eq 15 ]; then
+    echo "usage: $0 repo src_branch dst_branch dependent_k8s.io_repos required_packages kubernetes_remote subdirectory source_repo_org source_repo_name base_package is_library recursive_delete_pattern skip_tags last_published_upstream_hash remove_git_attributes"
```
liggitt, line 41: does recursive_delete_pattern not already support this?

nikhita:
Closing in favor of https://github.com/kubernetes/kubernetes/pull/108970

PR #293, opened 2022-05-11
Look in local cache first and then the normal proxy

dims:
Context : https://kubernetes.slack.com/archives/C2C40FMNF/p1652208676507389

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


nikhita (approved):
/lgtm
/approve

PR #312, opened 2023-02-20
update golangci-lint to v1.51.2

cpanato:
- update golangci-lint to v1.51.2

/assign @nikhita @dims


saschagrunert (commented):
Should we update the config as well?

cpanato:
> Should we update the config as well?

yes, we should but i was waiting for the action to run :(

can I do that in a follow up then we have the signal from ci as well

saschagrunert (approved):

PR #313, opened 2023-02-21
Fix lints and update golanci-lint config

cpanato:
- Fix lints and update golanci-lint config

### DIFF
--- a/cmd/update-rules/main.go
+++ b/cmd/update-rules/main.go
@@ -19,7 +19,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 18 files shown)


cmd/update-rules/main.go
```diff
@@ -126,7 +128,7 @@ func UpdateRules(rules *config.RepositoryRules, branch, goVer string) {
 		var newBranchRule config.BranchRule
 		// find the mainBranch rules
 		for _, br := range r.Branches {
-			if br.Name == "master" {
+			if br.Name == MainBranchName {
```
cpanato, line 131: As a follow up, should we prepare this to be both `master` and `main` to be ready when changing k/k branch name to `main`?

dims: sounds good as follow up PR.

PR #318, opened 2023-02-24
allow either master or main branch name

cpanato:
- allow either master or main branch name

/assign @dims @nikhita

### DIFF
--- a/cmd/update-rules/main.go
+++ b/cmd/update-rules/main.go
@@ -28,7 +28,10 @@ import (
 	"k8s.io/publishing-bot/cmd/publishing-bot/config"
 )
 
-const MainBranchName = "master"
+const (
+	MasterBranchName = "master"
+	MainBranchName   = "main"
+)
 
 type options struct {
 	branch    string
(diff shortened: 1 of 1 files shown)


palnabarun (approved):
/lgtm
/hold for others to take a look

cmd/update-rules/main.go
```diff
@@ -137,7 +140,7 @@ func UpdateRules(rules *config.RepositoryRules, branch, goVer string) {
 
 		// if mainBranch rules not found for repo, it means it's removed from master tree, log warning and skip updating the rules
 		if !mainBranchRuleFound {
-			glog.Warningf("%s branch rules not found for repo %s, skipping to update branch %s rules", MainBranchName, r.DestinationRepository, branch)
+			glog.Warningf("%s or %s branch rules not found for repo %s, skipping to update branch %s rules", MasterBranchName, MainBranchName, r.DestinationRepository, branch)
```
nikhita, line 143: The publishing-bot doesn't support the `main` branch yet.

What benefit do we get from having `update-rules` support `main`?

cpanato: To prepare when we rename the master branch to main in k/k 

if you think this makes no sense, we can close it, not a problem

nikhita: > if you think this makes no sense, we can close it, not a problem

I think this is a good addition! Let's fold it into a single PR which will add support for the `main` branch in the whole of the publishing-bot.

If we add it here now, we allow `main` in rules while it isn't really supported by the bot as a whole.

Closing for now but maybe @akhilerm can use this as a part of https://github.com/kubernetes/publishing-bot/pull/310 (containerd uses `main` so we'll need to add support for it at some point). Thanks for the PR though, @cpanato!! 😊 

/close

PR #362, opened 2023-07-09
repos.sh: add endpointslice repo

mrbobbytables:
ref: https://github.com/kubernetes/org/issues/4313

### DIFF
--- a/hack/repos.sh
+++ b/hack/repos.sh
@@ -35,6 +35,7 @@ repos=(
     cri-api
     csi-translation-lib
     dynamic-resource-allocation
+    endpointslice
     kms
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):

hack/repos.sh
```diff
@@ -35,6 +35,7 @@ repos=(
     cri-api
     csi-translation-lib
     dynamic-resource-allocation
+    endpointslice
```
akhilerm, line 38: @mrbobbytables Dont we need a corresponding change in the rules file in k/k also, so that the repo gets published.

PR #373, opened 2023-08-14
Fix fractional value memory limit

### LINKED ISSUE #365: Warning for memory limit of init container when deploying publishing bot
The default configurationfor the init container of the publishing bot, uses a fractional value ([1.6Gi](https://github.com/kubernetes/publishing-bot/blob/master/artifacts/manifests/podspec.yaml#L17)) for the memory limit.  This causes a warning when the  publishing-bot is deployed to a cluster
```
spec.template.spec.initContainers[0].resources.limits[memory]: fractional byte value "1717986918400m" is invalid, must be an integer
```

A valid memory limit value for the init container should be identified and used in the default configuration

ArkaSaha30:
Fixes #365 

This PR will change the memory limit from 1.6Gi to 1639Mi(_ceiling of `1638.4Mi`_) which does not represent a floating value.
1.6Gi = 1638.4 Mi = 1,677,721.6 Ki
1.5Gi = 1536 Mi = 1,572,864 Ki

### DIFF
--- a/artifacts/manifests/podspec.yaml
+++ b/artifacts/manifests/podspec.yaml
@@ -14,7 +14,7 @@ initContainers:
       memory: 200Mi
     limits:
       cpu: 2
-      memory: 1.6Gi
+      memory: 1639Mi
   volumeMounts:
   - mountPath: /etc/munge-config
     name: munge-config
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ INTERVAL ?= 86400
 CPU_LIMITS ?= 2
 CPU_REQUESTS ?= 300m
 MEMORY_REQUESTS ?= 200Mi
-MEMORY_LIMITS ?= 1.6Gi
+MEMORY_LIMITS ?= 1639Mi
 GOOS ?= linux
 
 build_cmd = GO111MODULE=on mkdir -p _output && GOOS=$(GOOS) go build -o _output/$(1) ./cmd/$(1)


ArkaSaha30:
cc @akhilerm

artifacts/manifests/podspec.yaml
```diff
@@ -14,7 +14,7 @@ initContainers:
       memory: 200Mi
     limits:
       cpu: 2
-      memory: 1.6Gi
+      memory: 1.5Gi
```
palnabarun, line 17: Even though the downward change is small, would it cause the init container to run into OOM errors? How about changing it into an exact value in `Mi`?

ArkaSaha30: Exact value in `Mi` = `1536 Mi` right?

akhilerm, line 17: I dont think we should reduce the memory limit. One option would be to add the value in `Mi` as suggested above.

I tried searching why the value `1.6Gi` was chosen but couldn't find a reason for it. We can also increase the limit of the initcontainer to be the same as the main container. WDYT. ?

akhilerm (commented):
we also need to update other occurrences of `1.6Gi` in the repository, so the bot can be run in default config also without any warning.

artifacts/manifests/podspec.yaml
palnabarun: Let's go with the equivalent of 1.6Gi = 1639Mi.

ArkaSaha30: Going with `1639Mi`. Adding the change in [Makefile](https://github.com/kubernetes/publishing-bot/blob/bb90a8b609b4fdfc44fb1fa7e1f74f7a2b3871f2/Makefile#L39) as well.

palnabarun (approved):
/approve

Leaving the lgtm to another reviewer.

artifacts/manifests/podspec.yaml
```diff
@@ -14,7 +14,7 @@ initContainers:
       memory: 200Mi
     limits:
       cpu: 2
-      memory: 1.6Gi
+      memory: 1639Mi
```
akhilerm, line 17: Can you also add a comment stating why this value is added. Users going through the spec should not get confused with a random number popping up for the memory limit.

PR #415, opened 2024-03-14
Prune unused replace directives

liggitt:
Fixes https://github.com/kubernetes/publishing-bot/issues/214

This is the same logic used to prune unused replace directives from the main kubernetes/kubernetes repo at https://github.com/kubernetes/kubernetes/blob/master/hack/update-vendor.sh#L355-L361

/hold for testing


thockin:
Manually verified

/lgtm
/approve

thockin:
This didn't seem to work:  https://github.com/kubernetes/kubernetes/issues/123933#issuecomment-2022763576

@dims @mohitsharma-in - I have no viz into how this is actually run - is the bot deployed with the latest commits?

liggitt:
@dims dug into logs and we found where the replace directives were getting added back. Followup fix in https://github.com/kubernetes/publishing-bot/pull/416

Issue #433, opened 2024-05-14
Enhance `rules.yaml` release branch validation to decouple from changes in k/k `master`


saschagrunert:
### Issue

We removed the `legacy-cloud-providers` staging repo in https://github.com/kubernetes/kubernetes/pull/124767 on k/k `master`, which immediately broke the rule validation ([`pull-publishing-bot-validate`](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/124829/pull-publishing-bot-validate/1789915699716558848)) for all supported release branches.

### Solution

- We only validate the rules for the actual release branch when running on a `release-x.y` branch. 
- Continue to validate all rules when running on `master`

cc @kubernetes/release-engineering

cpanato:
sounds good to me this approach

PR #434, opened 2024-05-14
Use `PULL_BASE_REF` for branch rules validation

saschagrunert:
We now consider the `PULL_BASE_REF` (which is set by prow but for testing can also be empty) to:

- Check all rules when running on `master`
- Check only branch specific rules for release branches

Fixes https://github.com/kubernetes/publishing-bot/issues/433

cc @kubernetes/release-engineering


cpanato (approved):
/lgtm

PR #438, opened 2024-06-12
Fix repeated corruption of packed-refs

dims:
We have seen this often enough that we should fix it programatically! most recent was today:
https://github.com/kubernetes/kubernetes/issues/56876#issuecomment-2162533093

We use the algorithm from the discussion we had with @soltysh and @sttts in:
https://kubernetes.slack.com/archives/C2C40FMNF/p1701175768598839?thread_ts=1700688672.734149&cid=C2C40FMNF

Today's discussion was in:
https://kubernetes.slack.com/archives/CHGFYJVAN/p1718172922410299


dims:
I've tested with the data we captured here:
https://storage.googleapis.com/k8s-infra-kops-ci-results/packed-refs.txt

artifacts/scripts/construct.sh
```diff
@@ -96,6 +96,32 @@ git config user.name "$GIT_COMMITTER_NAME"
 echo "Running garbage collection."
 git config gc.pruneExpire 3.days.ago
 git gc --auto
+
+# Remove corrupted lines in .git/packed-refs
+# Lines not starting with '^' caret are printed as-is. Lines
+# that do not start with a caret are printed ONLY if the previous
+# line did not start with a caret
+echo "Cleaning .git/packed-refs"
+cp packed-refs packed-refs.bak
```
upodroid, line 105: The file is inside the .git folder. Are the commands running inside that folder?

upodroid (changes requested):

upodroid:
Corrupted packed-refs file looks like this.

https://storage.googleapis.com/k8s-infra-kops-ci-results/packed-refs.txt

artifacts/scripts/construct.sh
```diff
@@ -96,6 +96,32 @@ git config user.name "$GIT_COMMITTER_NAME"
...
+    if (!prev_line_start_with_caret) {
+        print
+        prev_line_start_with_caret = 1
+    } else {
+        prev_line_start_with_caret = 1
+    }
+}
+' packed-refs.bak > packed-refs
+
 echo "Fetching from origin."
```
akhilerm, line 125: Should we run a `git gc --auto` after the cleaning up of packed-refs also?

artifacts/scripts/construct.sh
```diff
@@ -96,6 +96,32 @@ git config user.name "$GIT_COMMITTER_NAME"
...
+# Remove corrupted lines in .git/packed-refs
+# Lines not starting with '^' caret are printed as-is. Lines
+# that do not start with a caret are printed ONLY if the previous
+# line did not start with a caret
+echo "Cleaning .git/packed-refs"
+cp packed-refs packed-refs.bak
+awk '
+NR == 1 {
+    prev_line_start_with_caret = 0
+    print
```
akhilerm, line 109: This causes the first line (the comment `# pack-refs with: peeled fully-peeled sorted `  to be printed twice in the resulting file.

artifacts/scripts/construct.sh
```diff
@@ -96,6 +96,32 @@ git config user.name "$GIT_COMMITTER_NAME"
...
+}
+/^\^/ {
+    if (!prev_line_start_with_caret) {
+        print
+        prev_line_start_with_caret = 1
+    } else {
+        prev_line_start_with_caret = 1
+    }
+}
+' packed-refs.bak > packed-refs
```
akhilerm, line 123: The `packed-refs.bak` needs to be removed from the repo after performing the cleaning, else it may change the output of `git status`

artifacts/scripts/construct.sh
```diff
@@ -96,6 +96,32 @@ git config user.name "$GIT_COMMITTER_NAME"
...
+    prev_line_start_with_caret = 0
+    print
+}
+/^\^/ {
+    if (!prev_line_start_with_caret) {
+        print
+        prev_line_start_with_caret = 1
+    } else {
+        prev_line_start_with_caret = 1
+    }
```
akhilerm, line 121:
```suggestion
    if (!prev_line_start_with_caret) {
        print
    }
    prev_line_start_with_caret = 1
```

akhilerm (changes requested):
Tested with the packed-refs from client-go repo and a few other repos also, except the duplicated first line it works as expected.

artifacts/scripts/construct.sh
dims: thanks fixed

artifacts/scripts/construct.sh
dims: no should be fine from what i researched.

artifacts/scripts/construct.sh
dims: doing this!

artifacts/scripts/construct.sh
dims: fixed

artifacts/scripts/construct.sh
dims: whoops!

artifacts/scripts/construct.sh
akhilerm: I couldnt find where the `.git/packed-refs` was being removed. Since its not in the `.git` folder it wont affect the repo state, but prefer not having temporary files kept.

dims:
> Tested with the packed-refs from client-go repo and a few other repos also, except the duplicated first line it works as expected.

I removed the print statement in `NR == 1` did that not fix the duplicated line? can you please recheck?

akhilerm (approved):

akhilerm:
> I removed the print statement in `NR == 1` did that not fix the duplicated line? can you please recheck?

Yep, its working fine. The comment was part of the previous review.

